PR #3, opened 2012-11-06
Add STALE value to diagnostic_msgs/DiagnosticStatus/level enum

jbohren:
It appears it's become common to set diagnostic_msgs/DiagnosticStatus/level=3 to represent a "STALE" status.  Since this has become a convention, it makes sense to add this definition to the enum. 

See use of STALE status here: http://www.ros.org/doc/api/diagnostic_msgs/html/msg/DiagnosticStatus.html

See this answers thread for more info and discussion: http://answers.ros.org/question/47438/diagnostic_msgdiagnosticstatus-should-have-stale

### DIFF
--- a/diagnostic_msgs/msg/DiagnosticStatus.msg
+++ b/diagnostic_msgs/msg/DiagnosticStatus.msg
@@ -5,6 +5,7 @@
 byte OK=0
 byte WARN=1
 byte ERROR=2
+byte STALE=3
 
 byte level # level of operation enumerated above 
 string name # a description of the test/component reporting


jack-oquin:
Be careful. It is possible that people decided not to add it explicitly to avoid changing the message checksum. That could cause lots of trouble for people with large numbers of saved bags containing diagnostic messages (a reasonable practice).

Maybe the Stale status should just be added as a comment, leaving the checksum unmodified.

jbohren:
Do the MD5 checksums change if the const values change?

jbohren:
Yes, the checksum would definitely change: http://www.ros.org/wiki/ROS/Technical%20Overview#Message_serialization_and_msg_MD5_sums

So in that case, maybe we can stage this in for Hydro and just add the appropriate migration rules: http://ros.org/wiki/rosbag/migration

jack-oquin:
That's another option. 

This is kind of a nasty corner case. In general, changing or adding a const _could_ require message conversion. But, in this case existing messages would not actually be affected.

jack-oquin:
How was this resolved?

Looking at the code, Jonathan's patch does not appear to have been applied. But, I might be confused about that.

Do github issues have any kind of closing code, like the Trac WontFix?

dirk-thomas:
Oh, that seems to be a side effect of my recent action. I switched from "master" to "groovy-devel" which seems to automatically close the pending pull requests since the destination branch is gone. @jbohren could you please update (and reopen) the pull request to be against groovy-devel? Than it stays at least open and will be handled?

For issues you can attach labels (and most of the ROS repos should have a "wontfix" label). But this one should stay open until being resolved or really classified as wontfix.

jack-oquin:
OK, I was just wondering what the resolution was.

I personally favor adding the STALE const as a comment, rather than applying Jonathan's patch literally.

PR #4, opened 2012-11-19
Add sensor_msgs/MultiEchoLaserScan

chadrockey:
This message was never added after the review:
http://www.ros.org/wiki/sensor_msgs/Reviews/2012-08-01_MultiEchoLaser_API_review_API_Review

Final approval given in email by Jack O'Quin, Tully Foot, Christian Dornhege, and Eric Perko on August 30th.


chadrockey:
Fixed distance rangers was also approved.

Issue #6, opened 2013-04-24
sensor_msgs::convertPointCloudToPointCloud2() crashes


bmagyar:
I've been running experiments with the tabletop_object_detector and found that the segmentation crashed. Those detection/segmentation nodes are passing pointclouds by sensor_msgs::PointCloud, not PointCloud2. 
The bug always comes when sensor_msgs::convertPointCloudToPointCloud2() is called. I captured the point clouds that are being pushed to the converter at that time.

I set up a node that reproduces the bug. The code is the following:
http://pastebin.com/j6PUF2G7

Bag file with the data that produces the error: (~10MB)
https://www.dropbox.com/s/4fwc0d7d8vihsq0/points_cut2.bag

I managed to fix it by adding 'volatile' to 'int offset' in the function.

chadrockey:
In the future, it's better to immediately create a pull request of the code that works for you.  This automatically opens an issue related to that pull request and ensures that we see exactly what the fix is and that you remember the entire fix.  Can you please create a pull request for the fix and reference this issue there?

It would even better if you created a unit test from your example node that shows the issue.  Then we can make sure it always works.

PR #7, opened 2013-04-26
Added volatile to offset variable. 

bmagyar:
The crash was coming from optimization issues which is solved with volatile.
Possible source of this problem is a -ftrapv compilation flag I have.

Solution for issue:
https://github.com/ros/common_msgs/issues/6

Explanation copied from the above issue:
I've been running experiments with the tabletop_object_detector and found that the segmentation crashed. Those detection/segmentation nodes are passing pointclouds by sensor_msgs::PointCloud, not PointCloud2.
The bug always comes when sensor_msgs::convertPointCloudToPointCloud2() is called. I captured the point clouds that are being pushed to the converter at that time.

I set up a node that reproduces the bug. The code is the following:
http://pastebin.com/j6PUF2G7

Bag file with the data that produces the error: (~10MB)
https://www.dropbox.com/s/4fwc0d7d8vihsq0/points_cut2.bag

### DIFF
--- a/sensor_msgs/include/sensor_msgs/point_cloud_conversion.h
+++ b/sensor_msgs/include/sensor_msgs/point_cloud_conversion.h
@@ -74,7 +74,7 @@ static inline bool convertPointCloudToPointCloud2 (const sensor_msgs::PointCloud
   output.fields.resize (3 + input.channels.size ());
   // Convert x/y/z to fields
   output.fields[0].name = "x"; output.fields[1].name = "y"; output.fields[2].name = "z";
-  int offset = 0;
+  volatile int offset = 0;
   // All offsets are *4, as all field data types are float32
   for (size_t d = 0; d < output.fields.size (); ++d, offset += 4)
   {


wjwwood:
@bmagyar I can't reproduce this on OS X or Ubuntu Precise.

Can you explain a little why you think the optimization of the local `offset` variable causes your segfault? It seems to me that this might be resolving a side effect rather than the source of the problem.

What architecture are running on, i386, amd64, or armel/armhf?

Can we get a `VERBOSE=1` output of make (can be for the test case you proposed here) and the type and version of compiler you are using?

Sorry for the inquisition, I just want to make sure we solve the correct problem.

bmagyar:
Thanks for the feedback, I forgot to test on other machines :) 

These are the flags that I have on the desktop:
-O3 -std=c++0x -g -ggdb -mmmx -O3 -msse2 -msse3 -mssse3 -ftrapv -Wall -Wextra -Werror=address -Werror=array-bounds -Werror=c++0x-compat -Werror=char-subscripts -Werror=enum-compare -Werror=implicit-int -Werror=implicit-function-declaration -Werror=comment -Werror=conversion-null -Werror=div-by-zero -Werror=format-security -Werror=format-extra-args -Werror=init-self -Werror=int-to-pointer-cast -Werror=missing-braces -Werror=missing-field-initializers -Werror=return-type -Werror=nonnull -Werror=parentheses -Werror=pointer-arith -Werror=pointer-sign -Werror=reorder -Werror=return-type -Werror=sequence-point -Werror=strict-overflow=1 -Werror=trigraphs -Werror=type-limits -Werror=unused-value -Werror=volatile-register-var -fdiagnostics-show-option -Werror=overflow 

The system I'm running is i386, Ubuntu 12.04 32bit. 
I also tested it on my laptop (same config) which doesn't have these compilation flags set and it doesn't crash.

Probably -ftrapv, because it generates a nice and clean abort but I just cannot get how would that variable overflow for the size of pointclouds that I have:  307200 
Incrementation in that code is by 4 and it's an int so shouldn't cause overflow.

Hopefully I'll have time to set up an example for this in the next few days (without pointclouds).

dirk-thomas:
Closing for now. If you have further information please comment and the issue might be reopened.

PR #8, opened 2013-05-07
added action definition for getting maps from map_server

jonbinney:
Adds an action that does the same thing as the static_map service call, for people who want to use actionlib instead of services.

### DIFF
--- a/nav_msgs/CMakeLists.txt
+++ b/nav_msgs/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(nav_msgs)
 
-find_package(catkin REQUIRED COMPONENTS geometry_msgs message_generation std_msgs)
+find_package(catkin REQUIRED COMPONENTS geometry_msgs message_generation std_msgs actionlib_msgs)
 
 add_message_files(
   DIRECTORY msg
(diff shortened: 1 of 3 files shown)


hershwg:
This pull request looks good to me, but I'm not merging it right off the bat because of the recent email from @tfoote saying "Please hold on new merges for groovy so we can get a sync to public."

This is an addition of new a new .action file, so I don't expect it to cause any regressions.

dirk-thomas:
Why should the same operation which already exists as a service be available as an action if it does not have any progress message? That looks like duplicating API to me.

hershwg:
We have some infrastructure which doesn't support service calls or latched topics, so an action call seemed like a simple way to get the behavior we want through the supported pub/sub communications.

We are also under the impression that service calls are being discouraged in favor of topic-based solutions like actionlib.

dirk-thomas:
I assumed that service are always available when actionlib is. If that is not the case sure it makes sense to also expose the functionality as an action.

hershwg:
So... can I merge it?  Or would you like to?  Or do we need to wait for something else to happen?

dirk-thomas:
That should be the maintainers responsibility: @tfoote 
I was just asking questions out of curiosity.

tfoote:
@hershwg FYI the merge limitation is just for pulls into rosdistro not for individual packages.

PR #12, opened 2013-07-30
adding a JointCommands message (complement of JointState.msg)

hsu:
Proposal for setting joint commands.  Feedback welcome.

### DIFF
--- a/sensor_msgs/msg/JointCommands.msg
+++ b/sensor_msgs/msg/JointCommands.msg
@@ -1,0 +1,24 @@
+# Joint Command Message
+# This structure contains the gains to be applied to a joint.
+# The controller is a PID with feedforward desired torque:
+#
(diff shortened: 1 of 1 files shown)


chadrockey:
I don't think that commands belong in sensor_msgs.  Especially since this message has a lot of overlap with the discussion here: https://github.com/ros/common_msgs/pull/10

Should likely be with control_msgs or controller_msgs depending on which one survived - assuming this doesn't have significant redundancy with one of those messages.

chadrockey:
http://ros.org/wiki/control_msgs was the survivor.

isucan:
I really do not think this should go in sensor_msgs. I think this can go to control_msgs though, if there is no overlap with other messages.

hsu:
agreed, control_msgs is a better place for this.

PR #13, opened 2013-07-30
add effort for joint trajectories

isucan:
After a discussion on the moveit mailing list, we think it would be best to add effort[] to JoinTrajectory. 
Attn: @sachinchitta @mikeferguson @shaun-edwards @davetcoleman

### DIFF
--- a/trajectory_msgs/CMakeLists.txt
+++ b/trajectory_msgs/CMakeLists.txt
@@ -15,3 +15,6 @@ add_message_files(
 generate_messages(DEPENDENCIES std_msgs geometry_msgs)
 
 catkin_package(CATKIN_DEPENDS message_runtime std_msgs geometry_msgs)
+
+install(DIRECTORY rules
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
--- a/trajectory_msgs/msg/JointTrajectoryPoint.msg
+++ b/trajectory_msgs/msg/JointTrajectoryPoint.msg
@@ -1,4 +1,5 @@
 float64[] positions
 float64[] velocities
 float64[] accelerations
+float64[] effort
 duration time_from_start
--- a/trajectory_msgs/msg/JointTrajectoryPoint.msg
+++ b/trajectory_msgs/msg/JointTrajectoryPoint.msg
@@ -1,3 +1,7 @@
+# Each trajectory point specifies either position[, velocities[, accelerations]]
+# or positions[, effort] for the trajectory to be executed.
+# All specified values are in the same order as the joint names in JointTrajectory.msg
+
 float64[] positions
 float64[] velocities
 float64[] accelerations
--- a/trajectory_msgs/msg/JointTrajectoryPoint.msg
+++ b/trajectory_msgs/msg/JointTrajectoryPoint.msg
@@ -1,4 +1,4 @@
-# Each trajectory point specifies either position[, velocities[, accelerations]]
+# Each trajectory point specifies either positions[, velocities[, accelerations]]
 # or positions[, effort] for the trajectory to be executed.
 # All specified values are in the same order as the joint names in JointTrajectory.msg
 
--- a/trajectory_msgs/package.xml
+++ b/trajectory_msgs/package.xml
@@ -19,4 +19,10 @@
   <run_depend>message_runtime</run_depend>
   <run_depend>geometry_msgs</run_depend>
   <run_depend>std_msgs</run_depend>
+  <run_depend>rosbag_migration</run_depend>
+
+  <export>
+    <rosbag_migration_rule rule_file="rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr" />
+  </export>
+
 </package>
--- a/trajectory_msgs/package.xml
+++ b/trajectory_msgs/package.xml
@@ -19,7 +19,7 @@
   <run_depend>message_runtime</run_depend>
   <run_depend>geometry_msgs</run_depend>
   <run_depend>std_msgs</run_depend>
-  <run_depend>rosbag_migration</run_depend>
+  <run_depend>rosbag_migration_rule</run_depend>
 
   <export>
     <rosbag_migration_rule rule_file="rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr" />
--- a/trajectory_msgs/rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr
+++ b/trajectory_msgs/rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr
@@ -1,14 +1,14 @@
 class update_trajectory_msgs_JointTrajectoryPoint_84fd2dcf68773c3dc0e9db894f4e8b40(MessageUpdateRule):
-    old_type = "trajectory_msgs/JointTrajectoryPoint"
-    old_full_text = """
+	old_type = "trajectory_msgs/JointTrajectoryPoint"
+	old_full_text = """
 float64[] positions
 float64[] velocities
 float64[] accelerations
 duration time_from_start
 """
 
-    new_type = "trajectory_msgs/JointTrajectoryPoint"
-    new_full_text = """
+	new_type = "trajectory_msgs/JointTrajectoryPoint"
+	new_full_text = """
 # Each trajectory point specifies either positions[, velocities[, accelerations]]
 # or positions[, effort] for the trajectory to be executed.
 # All specified values are in the same order as the joint names in JointTrajectory.msg
@@ -20,15 +20,15 @@ float64[] effort
 duration time_from_start
 """
 
-    order = 0
-    migrated_types = []
+	order = 0
+	migrated_types = []
 
-    valid = True
+	valid = True
 
-    def update(self, old_msg, new_msg):
-        new_msg.positions = old_msg.positions
-        new_msg.velocities = old_msg.velocities
-        new_msg.accelerations = old_msg.accelerations
-        #No matching field name in old message
-        new_msg.effort = []
-        new_msg.time_from_start = old_msg.time_from_start
+	def update(self, old_msg, new_msg):
+		new_msg.positions = old_msg.positions
+		new_msg.velocities = old_msg.velocities
+		new_msg.accelerations = old_msg.accelerations
+		#No matching field name in old message
+		new_msg.effort = []
+		new_msg.time_from_start = old_msg.time_from_start
--- a/trajectory_msgs/rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr
+++ b/trajectory_msgs/rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr
@@ -1,0 +1,34 @@
+class update_trajectory_msgs_JointTrajectoryPoint_84fd2dcf68773c3dc0e9db894f4e8b40(MessageUpdateRule):
+    old_type = "trajectory_msgs/JointTrajectoryPoint"
+    old_full_text = """
+float64[] positions
+float64[] velocities
+float64[] accelerations
+duration time_from_start
+"""
+
+    new_type = "trajectory_msgs/JointTrajectoryPoint"
+    new_full_text = """
+# Each trajectory point specifies either positions[, velocities[, accelerations]]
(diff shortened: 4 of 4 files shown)


davetcoleman:
I recall some of the discussion from the moveit mailing list, but basically we are adding for special situations such as grippers?

Typically a 7dof arm joint trajectory will not contain efforts, however, as that is usually calculated by the trajectory controllers, right? A vector of 0 valued efforts can typically be considered as ignored, right? How does a controller know the difference? Is there a mode flag or something?

I think adding a few comments to this message would help clear this up for future users. Having values for positions, velocities, accelerations AND efforts seems like too much information at first glance.

isucan:
Yes, this is for special situations like grippers (but perhaps not only; I guess it depends on what controllers you plan to use). The way I think about it is that whoever publishes this message will include either positions,[ velocities, [accelerations]] _or_ positions, [effort]. (`[ ]` denotes optional arguments). When not specified, the effort will be simply a vector of length 0, so there will be no values there. When effort is not specified, there is no confusion. A confusion is possible if velocities accelerations and effort are specified. In that case, we should probably give a warning and ignore effort.
The other option is to add a new trajectory message that has positions & effort, but that seems a bit overkill.

davetcoleman:
Makes sense and I +1 this pull request, I just think you should add basically what you just said as a comment in the message

sachinchitta:
Looks good to me.

mikeferguson:
One thing to note is that this is a pretty low level message, used by a large number of packages. By adding a field, we end up changing the MD5. Luckily it is not something is commonly bagged up and kept around as a dataset (as is the case with things like images and point clouds).

This should certainly only go into hydro only (it's far too late to make a change like this in groovy) and should definitely be noted in the migration guide.

davetcoleman:
Good point. Robots like Baxter will be on Groovy for the foreseeable future (year?), and this will break Hydro-Groovy compatibility :-/

isucan:
That is a good point. If someone wants to run < hydro on the robot and publish trajectories, an adapter will be needed. This is not fun... but this way we will never change the message...

shaun-edwards:
+1 from me.  There are a lot of robots that can be controlled with joint
effort and this will open up that level of control.

mikeferguson:
@davetcoleman @isucan Yeah, you would have to do a republishing adapter for any robot that is running groovy against a hydro moveit -- although, I'm pretty certain there are probably plenty of other incompatibilities that nobody bothered to document, and so in my experience trying to run two different distros and hook them together is generally a painful time and a bad idea.

The way I look at this, either we fix the underlying JointTrajectoryPoint, or we will probably end up defining a new JointTrajectoryPoint2 and JointTrajectory2, use those in MoveIt, and then instead of the occasional republishing for robots that are running Groovy but connecting to Hydro, everyone who wants to connect to MoveIt ends up having to change the name of the message in every driver/script/program they've written.

I took a look at the 'used by' list for trajectory_msgs, of course this is only directly depending packages and only packages that are released, but the list is comprised mainly of moveit/arm_nav, things I've written, the ROS-Industrial stuff, and teleop/pr2-demo packages. I don't see anything in the list that screams "I'm saving lots of these messages to bagfiles and I will hate you all if you go and change the MD5 checksum of it"

Even though we're a bit past the "feature freeze" for hydro and shouldn't really be doing this, I'll give it a +1 cause the alternatives suck even more.

tfoote:
As Mike pointed out, this is really late in the hydro cycle for a change to a core message.  

If this is to be merged we need to have a full analysis of who it's going to effect, what they need to do to migrate and what the likelyhood of this change breaking things is.  To the level of what type of code changes are going to be necessary.  With this being an optional additional field, it probably won't be too bad, but this needs to be carefully reviewed that we won't break any common use cases.  A quick grep shows some 200+ usages of this message in released code.  

Also, as this is a change we will need a If we were to do this we will need a rosbag migration rule and test for the migration

shaun-edwards:
Is there a specific need to make this change in Hydro?  I realize with
longer release cycles, we probably want to get this change in there sooner
rather than later.  There has some talk of and LTS for I-turtle, so we
definitely want to get this into ROS before then.

mikeferguson:
@tfoote Given how these messages are used, I don't think any code needs to be updated (basically, all existing code is simply going to ignore the effort, even if it is filled in and the outputs of those programs will also be ok, because an empty vector of efforts is perfectly acceptable).

The rosbag migration therefore is the big issue, and does need to be addressed. IMO, with a proper rosbag migration, this change should not cause any real headaches -- and certainly will cause less headaches than if we went and defined a new message started using that (which is basically the alternative), as we saw with the slow move from pr2_controllers/FollowJointTrajectoryAction to control_msgs/FollowJointTrajectoryAction.

chadrockey:
Just to check, it looks like you can use either the new way (p + effort) or the old way (p + v + a).  Since this is the case, do we expect subscribers to dynamically switch types (p + e) until a grasp, then (p + v + a) to release?  If joints are not expected to change modes, does it make sense to make these the same message?  @isucan was talking about being able to populate the message in a conflicting way...

davetcoleman:
Given the delay until the next ROS release and the seemingly important need for effort in gripper joint commands I +1 doing this now despite the consequences. Its a minor change except for the md5 difference.

isucan:
@chadrockey I think the way the message is defined, you cannot switch from p +e  to p + v+ a in one trajectory -- you would need two messages. Do you think a separate message (JointTrajectoryEffort.msg) that is just p + e  would make more sense? (and then leave this one as it is).

sachinchitta:
We need to get this through in Hydro so we can develop against it going forward. Otherwise, the wait is too long for Indigo. Also, we should not be introducing yet another message into the mix - its fairly easy to check a message and figure out what type it is - some of our controllers already do that.

mikeferguson:
@chadrockey Even in the old message, you had options of (p+v+a), (p+v), heck even just-p is valid, and although nobody probably used it, you could have specified (v+a) or even just (v). It really all depends on the planner/smoother being used, and on the end, you find the code typically checks what is available and ignores the message with an error if something is not right (for instance, most of the PR2 controllers fail if you don't have at least P+V, and will choose either cubic or quintic splines depending on whether accelerations are filled in).

chadrockey:
My only concern was a message that can contradict itself and will have multiple but correct behaviors based on how the controller interprets it.  (Controller A works for a year using effort; switch to Controller B - uses accelerations instead: bad things happen.)

@isucan My question was if it's expected that a joint will execute one entire trajectory in (pva), then continue on another trajectory in (pe).

Ultimately, this change should be accepted if control modes will switch in this manner.

If they do not switch in this manner, it should be JointTrajectoryEffort.msg - which avoids a migration, can be backported for Groovy robots, and cannot contain an invalid/ambiguous trajectory.

sachinchitta:
The expectation is that all points in a trajectory message will have to be executed in the same control mode - i.e. you won't be mixing an effort only trajectory with a position only trajectory in the same message.

mikeferguson:
@tfoote, where do we want to put migration rules? Do we want to include it in the repo with an export?

I've generated the migration rule for the JointTrajectoryPoint:

```
class update_trajectory_msgs_JointTrajectoryPoint_84fd2dcf68773c3dc0e9db894f4e8b40(MessageUpdateRule):
    old_type = "trajectory_msgs/JointTrajectoryPoint"
    old_full_text = """
float64[] positions
float64[] velocities
float64[] accelerations
duration time_from_start
"""

    new_type = "trajectory_msgs/JointTrajectoryPoint"
    new_full_text = """
# Each trajectory point specifies either positions[, velocities[, accelerations]]
# or positions[, effort] for the trajectory to be executed.
# All specified values are in the same order as the joint names in JointTrajectory.msg

float64[] positions
float64[] velocities
float64[] accelerations
float64[] effort
duration time_from_start
"""

    order = 0
    migrated_types = []

    valid = True

    def update(self, old_msg, new_msg):
        new_msg.positions = old_msg.positions
        new_msg.velocities = old_msg.velocities
        new_msg.accelerations = old_msg.accelerations
        #No matching field name in old message
        new_msg.effort = []
        new_msg.time_from_start = old_msg.time_from_start
```

@isucan If it turns out we do want to put this right into the repo, the details are here: http://ros.org/wiki/rosbag/migration, but basically we would put this in a file called something like "jointtrajectorypoint.bmr" and add the following to the manifest (I assume this still works with catkin): 

```
<depend package="rosbag"/>
<export>
  <rosbag migration_rule_file="jointtrajectorypoint.bmr"/>
</export>
```

I'm not sure how I feel about common_msgs depending on rosbag, that feels wrong, so please do point out where this file should go.

isucan:
bump

tfoote:
We should not add a dependency on rosbag for the migration at that level.  Probably we should add the migration rule into something like rosbag itself so that if you have rosbag you get the migration rules.

isucan:
I put the rule files into these commits:
https://github.com/isucan/ros_comm/commit/c234aac5c47363eac6fcfbebb8c20d69774cceef
https://github.com/isucan/ros_comm/commit/38b0e578ba450897b56b7fc876733254f901666d
I plan to send a pull request as soon as the hydro-devel branch gets created for ros_comm.

isucan:
@tfoote @dirk-thomas -- have you thought about the pull request on ros_comm? If possible to merge this, I'd like to push some corresponding changes to moveit. This really is a very minor change on the message itself (just an addition) that should not affect existing code.

dirk-thomas:
We talked about it and putting rosbag migration rules into the _rosbag_ is not desired. The rosbag wiki page (http://ros.org/wiki/rosbag/migration) explicitly recommends to keep the message files and their rules together.

Obviously a dependency on rosbag is not desired therefore we should create a package _rosbag_migration_ package which will basically be empty and serves as an _interface_ to export rule files. _trajectory_msgs_ then depends on that package and exports the rule file which it stores beside the modified message file. _rosbag_ will then use these exports beside the already supported one.

Does that sound like a viable solution for your case?

isucan:
This is fine for me. Where would this rosbag_migration package live?

dirk-thomas:
Beside rosbag in the ros_comm repo.

dirk-thomas:
Can you please try ros/ros_comm#273, update this pull request to contain the rule file and export it with:

```
<export>
    <rosbag_migration_rule rule_file="rules/JointTrajectoryPoint_Groovy2Hydro_08.10.2013.bmr" />
</export>
```

dirk-thomas:
Please check this soon in order for the patch to be released in the upcoming ros_comm release.

isucan:
@dirk-thomas The rule now works. I was able to migrate messages properly using it and I believe it can be merged. https://github.com/ros/ros_comm/issues/273 can also be merged.

There is an issue that I ran into with rosdep2 (probably better suited for a separate ticket, but if anyone wants to duplicate my testing, this info currently needed):

Since rosbag_migration_rule is just local, it seems not to be detected as a ROS package by rosdep2.rospack.is_ros_package().
So I had to change that from

```
def is_ros_package(view, rosdep_name):
    return _ros_flag(view, rosdep_name, True)
```

to

```
def is_ros_package(view, rosdep_name):
    if rosdep_name == "rosbag_migration_rule":
        if not _ros_flag(view, rosdep_name, True):
            print ("FAIL")
            return True
    return _ros_flag(view, rosdep_name, True)
```

This in turn makes the manifest parser to ignore the dependency on rosbag_migration_rule and then the export rule is not found:

```
if is_ros_package(_static_rosdep_view, d.name):
                     depends.add(d.name)
if is_system_dependency(_static_rosdep_view, d.name):
                     rosdeps.add(d.name)
# both if statements above return false
```

I did not dig deeper into finding why the problem occurs once I verified the migration rule works. Things will work fine once rosbag_migration_rule is released, but it'd be great if this function were fixed too.

tfoote:
Ok, thanks everyone for the work on this.  The new rosbag_migration_rules is being built on the farm.  Once it's configured properly I'll merge and release this.

UltronDestroyer:
Sorry to revive this folks, I wanted to confirm that the conclusion from this discussion means that any code using the Groovy definition of trajectory_msgs/JointTrajectoryPoint.msg must be modified to include a .effort value to prevent MD5 errors?

mikeferguson:
@TheDash when moving to newer releases you don't need to do anything with your code (leaving the effort vector of length 0 should tell all subscribers that the effort is not being used/defined), but you can't mix groovy and hydro components that are using trajectory_msgs

PR #14, opened 2013-09-11
line wrap formating

dawonn:
http://docs.ros.org/api/sensor_msgs/html/msg/Imu.html

The webdoc doesn't word wrap, so I was wondering if you thought this is a good idea or not?


tfoote:
cherry-picked forward into hydro-devel fabb54d

PR #16, opened 2013-10-31
Add a C++ PointCloud2 Iterator

vrabaud:
Here are the packages that use PCL for PointCloud2 conversion.
- [ ] navigation
- [x] octomap_mapping
- [ ] velodyne
- [ ] velodyne_height_map

The following packages could even drop the PCL dependency:
- [x] calibration
- [x] gazebo_ros_pkgs
- [x] image_pipeline
- [x] moveit_ros
- [x] octomap_ros

If we add a TransformPointCloud method, more than half of the first set can also drop the PCL dependency.


vrabaud:
Maybe @ahornung is interested in that too (for Octomap ROS to maybe only depend on pure ROS stuff)

ahornung:
Yeah, minimizing dependencies is always great especially for such lightweight stuff! :+1:

tfoote:
This looks great.  I'd like a few people to review the API.

Recent helpful reviewers have been
@wjwwood @chadrockey @hershwg @mikeferguson

Could you guys take a look and comment?

mikeferguson:
adding @jack-oquin as well

mikeferguson:
As we've recently been looking at the ROS CppStyleGuide (during discussions on the linter), I'm going to raise that passing by reference an item which is non-const and is then modified is in contradiction to the guide (http://wiki.ros.org/CppStyleGuide, see section 10), and is used in the two SetFields methods.

Otherwise, seems fine to me, and certainly a good way to get rid of some heavier dependencies when just loading data.

vrabaud:
@mikeferguson  , thx, I fixed that. Something I should change is the readability: I can put some implementations in a private header. Do we have a convention for that ? like point_cloud_iterator_p.h or a private folder ?

jack-oquin:
@vrabaud: I suggest putting the non-public implementation details in an `impl` subdirectory:

```
include/sensor_msgs/point_cloud_iterator.h
include/sensor_msgs/impl/point_cloud_iterator.h
```

Of course, nothing keeps people from using those details except common sense, but it imposes none of the overhead of a PIMPL implementation, which would be unnecessary and in appropriate in this case.

vrabaud:
@jack-oquin  , will do. There is more than common sense to protect people from using those details: the anonymous namespaces they're in :) My goal is just to have less code to make the usage self-explanatory. On it.

vrabaud:
@jack-oquin  I made the implementation hidden.
Now, all, I have several questions:
- do you think the naming is right ? (setPointCloud2FieldsByString feels long)
- should the setPointCloud2Fields take a PointCloud2::Ptr or a PointCloud2\* ? (more generic with the second option but maybe less elegant). Overloading probably is the right solution.
- should I include other types in setPointCloud2FieldsByString ? (normals ?)
  Thx.

jack-oquin:
Vincent:

Thanks for doing this. It will be quite useful. Several general comments:
- At first, I had trouble figuring out exactly what it was. I got confused, thinking you were iterating over the list of fields in the PointCloud2 message, rather than the data for a list of fields. Perhaps an introductory comment at the file level would help orient the reader with a usage example or two. Most specific interfaces already have good examples, which helps. 
- I agree that a shorter and clearer name for `setPointCloud2FieldsByString()` would help. That name suggests it takes an arbitrary list of field names, which is actually what `setPointCloud2Fields()` does (I believe). The "Strings" mentioned above are described as: "xyz" (3 floats), "rgb" (3 uchar stacked in a float), "rgba" (4 uchar stacked in a float). Are those the only options? Is there some other word that better describes these useful groups of fields? Would it help to provide specific or overloaded interfaces for common combinations like "XYZ", "XYZRGB", "XYZRGBA"? 
- I am a uncomfortable with the use of varargs, but don't have a better idea to suggest. 
- Naive question: do C++ iterators usually provide `operator<=()`?

wjwwood:
> Perhaps an introductory comment at the file level would help orient the reader with a usage example or two.

I would like to see some simple scenarios/examples in the documentation for the file, a sort of what's in this file and how could I use it type of thing.

vrabaud:
@jack-oquin  and @wjwwood , I improved/fixed the docs so that should help.
@jack-oquin , I'd like to keep the PointField strings close to what they are in the end ("x", "y", "z" and "rgb") so no upper case or too many string concatenation (like "xyzrgb"). To be strict, I could only allow "x", "y", "z" and not "xyz". Anybody else for that ?
C++ iterators implement more things than what I did (like "operator <=" or "operator -). I wanted to keep it simple while providing all that a user would need. If you think <= is compelling, I'll add it to the list.

New question all: should I split setPointCloud2Fields and the iterator in two different files ? (point_cloud2_iterator.h and point_cloud2_utils.h).

vrabaud:
ok, I pushed a new round of modifications: I kept the old history for now even if my new code is different. I kept the function to change fields but instead of having an iterator, I created a proxy class through which a sensor_msgs::PointCloud2 can be read/modified. With that file, I checked all the packages that directly depend on pcl_msgs, pcl_ros or pcl_conversion and I was able to remove such dependency (like in most of navigation, moveit, image_pipeline, gazebo_plugins). In the end, only calibration, laser_filters, velodyne, costmap_2d and base_local_planner still need PCL, but only for the transformPointCloud function (which is just a for loop and that could be rewritten from scratch of maybe included in my PR).

Two suggestions to move forward:
- have the PointXYZ and PointXYZRGB be members of the proxy classes to not pollute the sensor_msgs namespace
- have a macro to "register" a new Point type so that setFields can be removed and done internally through the Proxy (a bit like done in PCL but much simpler as we don't need to deal with alignment). Right now, I modified velodyne and moveit (that both have custom Point types) calling setFields manually.

vrabaud:
and btw, for example, this is what the moveit code looks like after:
https://github.com/vrabaud/moveit_ros/commit/2d5606164bf779b85c1515aa3e7bf0df43bb3e3f
(I did not try to optimize anything, I just modified the code to be as close to the original as possible: as you can see, it's almost the same except for the conversions to/from PCL)

jack-oquin:
The moveit patch looks like an improvement to me.

vrabaud:
Here are other examples if needed:
https://github.com/vrabaud/image_pipeline/commit/98eac67065e26765855880f2a53c8de737462019
https://github.com/vrabaud/navigation/commit/4ab3eb84cb727b8c3a19107475d68a7041cfe0a9
Very few changes and no heavy PCL dependency anymore.

tfoote:
This is looking good and provides a nice compact API to the message.

ahornung:
Yes, looking good!

vrabaud:
I'm not in a hurry to merge that but I was wondering whether there was anything to fix in there. On what I suggested before:
- PointXYZ cannot be member classes as otherwise you could not template on it
- addPointField\* could be part of the container but it would be weird to template on some point type, and then use that function that defines the point type .... Plus that would make the container even more different from an STL container.

Actually, I can see a last issue: endianness. In PCL, it seems to be "solved" with a "/// @todo msg.is_bigendian = ?;" in conversions.h. Would something like that work ? (not an expert in endianness)

```
#if defined(BOOST_BIG_ENDIAN)
struct PointXYZ {
  float z;
  float y;
  float x;
};
#else 
struct PointXYZ {
  float x;
  float y;
  float z;
};
#endif
```

mikeferguson:
So, while investigating something completely unrelated, @dbking77 and I came across the fact that PCL-generated PointCloud2 messages, such as those put out by the kinect drivers, do some crazy padding (xyzrgb takes 32 bytes instead of 16bytes per point, see a discussion here http://www.pcl-users.org/PointXYZRGB-size-td2632598.html). 

@vrabaud it appears this PR would give us a way to publish xyzrgb taking only 16 bytes per point, but have you tried reading those clouds in other tools like RVIZ and/or something that uses the PCL conversions stuff? Does it happen to be smart enough to do it with SSE or does it die?

vrabaud:
Right, that is a bug and I had not seen it as RViz does not use PCL and I had not used the PCL conversions with my code (as it duplicates its functionality). Now, while writing a test for my fix, that got me thinking on the two solutions: the field iterator (my original proposition) and the container proxy (my second proposition).

The container proxy (which is also roughly what PCL provides now) has many disadvantages:
- it ties the message to some memory mapping (because of padding and endianness) which causes each point in the message to take 8 floats (3 for XYZ + 1 for padding + 1 for RGB + 3 for padding=32 bytes) instead of the 15 bytes of information it actually contains
- it ties us to how PCL structures its point type (which could change like in the PCL 2.0 drafts I had seen back in the days) and is therefore not future proof
- it assumes you can map the set of fields to a struct type at compile time (which PCL usually provides but what if you receive an unknown point type where all you care about is just X,Y,Z ?) and one of the main consequences I see around is that people just convert blindly a PointCloud2 msg to a pcl::PointCloudpcl::PointXYZ without even checking if those are the only fields present (pcl_conversion and PCL just trusts the user on that)
- it does not deal with endiannes (as you would need to check that at runtime and choose a PointXYZBigEndian or a PointXYZLittleEndian), which is not the case too actually in PCL from what I saw.

The only good thing about it is that it is convenient and fast as you just have a point type to deal (which can help for read/write or if you use a container, like in PCL).

This solution (and therefore what we actually use now through pcl_conversion) is NOT flexible, NOT future proof, wastes network resources and does not allow to write code that can work with different kinds of point clouds: I therefore revert my commits.

The only doable thing is the previous iterator I was proposing: it solves all the problems above but creates the following problems:
- it requires more code to be changed (as there is no nice point struct to use) but the overall amount of lines does not change so we're good
- it slows down reading/writing as it requires you to have an iterator per field but this conversuib time is most likely very small compared to the point cloud processing

Here is a roadmap I propose:
- I fix all code using pcl_conversion for message reading/writing using the field iterator (I had done it with the container proxy so it will be easy) and make sure my patch cannot be improved
- the PR is accepted in Hydro and in packages that could benefit from that
- in Indigo, we have to impose that people use that solution when dealing with a PointCloud2 they subscribe to
- in Indigo or J*, we publish smaller PointCloud2 messages and the iterator can magically understand old and new PointCloud2 fields thus offering BC

Thoughts ? (long but that was for the record, if there is a better mailing list, please let me know).

vrabaud:
any news on that ? I only got private positive feedback on my ROS users announcement. Do you guys want more tests ? Would be cool to have it for Indigo.

jack-oquin:
Vincent: I think this whole issue is a good idea, and something I personally want to use.

The questions you raise in your long message above are beyond my knowledge of the subject. Since I don't have time to delve into it in adequate detail, I don't feel qualified to comment.

tfoote:
So far I've only seen positive feedback on this well commented and circulated pull request. I will be releasing common_msgs shortly for indigo. I will plan to merge this once I fork development.

mikeferguson:
I concur that we should merge this into Indigo. Since the core/buildsystem is pretty stable, we should have ample time to actually test things before Indigo is released.

vrabaud:
This can also be merged into hydro as it's an independent header. People could then start porting during Hydro (by accepting my PR that are ready to go) which is good because some packages are pretty high up in the dependency tree and will not make it too soon to indigo.

mikeferguson:
I'm against merging this into Hydro, as it encourages people to make lots of changes in a supposedly "stable" system. Given that we know code currently using PCL to "decode" point clouds may have problems, we're going to end up with weird interaction issues between updated publishers and non-updated subscribers.

I expect all hell to break loose when I update to Indigo the first time, I do not expect all hell to break loose when I update my hydro debs to get a few recent patches. Obviously, if I am depending on something like MoveIt, which is still under development, I expect all hell to break loose every time I update, but merging this iterator into Hydro this late in the game really will promote "stable" packages getting changed.

vrabaud:
@mikeferguson  makes sense. It's true the coding/decoding is unstable enough already. Indigo it is then.

mikeferguson:
@vrabaud, so we don't miss anything, can you update the original description of this PR to include a checklist of the packages that we plan to update once this is merged? I know we've discussed a few already, and that you did a pretty exhaustive search of who was using PCL just for publish/subscribe.

tfoote:
merged into indigo 4fa63b931e64856fd2074d29b4e036c996ee7cee

tfoote:
@vrabaud Currently the unit tests are segfaulting.

tfoote:
@vrabaud n/m fixed 24a533d39e6e45642726110e5b98f32a598c5f64

vrabaud:
duuuuuuuuuuh. Thx. I guess I forgot to re-ran the tests then.

dimatura:
Hello, not sure if this is the right place but I gave this feature a try with the lastest indigo code (commit 48cddaa) and found a bug related to the use of `va_arg` inside the argument list of a function call. I made a pull request with a fix (9d3d7ba).

tfoote:
Closing. This was merged. The va_args was also merged. ( #37 )

tfoote:
I noticed that we still had the checklists for recommended updates. We should open tickets against those repos and link them to this ticket and then close this one.

vrabaud:
Updated the bug. I need to add the PointCloud transform to tf2, we can them properly remove PCL from those packages.

mikeferguson:
Ticketed against navigation -- https://github.com/ros-planning/navigation/issues/264

@vrabuad was there ever any progress on the transform stuff?

vrabaud:
actually yes, let me push that (I also tried to remove PCL from navigation but that required moving it to tf2 and that is biiiiiiig).

jack-oquin:
What we need is a rosdep for just the pcl-common components. That would be small, and lots of packages only require those basic definitions for manipulating point clouds.

mikeferguson:
@vrabaud I happened to compile with -Wall recently (after using the new tf2_sensor_msgs stuff, and came across the following warning related to PointCloud2Iterator:

```
/opt/ros/indigo/include/sensor_msgs/point_cloud2_iterator.h: In instantiation of ‘class sensor_msgs::PointCloud2Iterator<float>’:
/opt/ros/indigo/include/tf2_sensor_msgs/tf2_sensor_msgs.h:68:49:   required from here
/opt/ros/indigo/include/sensor_msgs/point_cloud2_iterator.h:282:7: warning: ‘sensor_msgs::PointCloud2Iterator<float>’ has a field ‘sensor_msgs::PointCloud2Iterator<float>::<anonymous>’ whose type uses the anonymous namespace [enabled by default]
 class PointCloud2Iterator : public PointCloud2IteratorBase<T, T, unsigned char, sensor_msgs::PointCloud2, PointCloud2Iterator>
       ^
```

YoshuaNava:
This is a very cool implementation of an iterator, thank you very much :+1: 

I have recently found some issues, which I describe in https://github.com/ros/common_msgs/pull/16

I'm writing here to be able to get in touch with the authors of the iterator, because I opened an issue in the ros2 repo.

PR #25, opened 2014-01-26
make link to control messages a real link

mikeferguson:
This is based on how common_msgs links, so that it gets rendered nicely on the wiki.


tfoote:
Thanks, looks good.

PR #27, opened 2014-02-05
find_package(actionlib_msgs) mangles catkin_LIBRARIES variable

dirk-thomas:
Derived from ros/geometry#51:

The Hydro `actionlib_msgs` package mangles the catkin_LIBRARIES variable if it is included with `find_package(actionlib_msgs)`. The CMake extra file must not depend on catkin: https://github.com/ros/common_msgs/blob/hydro-devel/actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em#L2

### DIFF
--- a/actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
+++ b/actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
@@ -1,5 +1,5 @@
 # need genmsg for _prepend_path()
-find_package(catkin REQUIRED COMPONENTS genmsg)
+find_package(genmsg REQUIRED)
 
 include(CMakeParseArguments)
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@roehling fyi

dirk-thomas:
@tfoote Please verify that the patch works and fixes the problem. I have only written in without running any configure step.

roehling:
The mangling is fixed, but I get a few `generate_messages` errors like this:

```
CMake Error at /home/roehling/ros/build/common_msgs/geometry_msgs/cmake/geometry_msgs-genmsg.cmake:167 (add_custom_target):
  add_custom_target cannot create target
  "geometry_msgs_generate_messages_cpp" because another target with the same
  name already exists.  The existing target is a custom target created in
  source directory "/home/roehling/ros/src/bug".  See documentation for
  policy CMP0002 for more details.
Call Stack (most recent call first):
  /opt/ros/hydro/share/genmsg/cmake/genmsg-extras.cmake:303 (include)
  common_msgs/geometry_msgs/CMakeLists.txt:33 (generate_messages)
```

My ROS workspace contains nothing but the `common_msgs` clone and the `bug` package.

roehling:
Here is the full build log: https://gist.github.com/roehling/8823026

roehling:
I just tried it without the fix, and the build error still occurs, so I'd say it is a problem with my build configuration.

dirk-thomas:
You are running a partial overlay workspace (like ros/genmsg#39). Rename `bug` to `z` as a temporary workaround for the duplicate targets.

roehling:
Thanks, that helped.

tfoote:
Confirmed the patch fixes the test code.

tfoote:
cherry-picked back to groovy-devel 394288cd624670d300800887acf100d06d734862

Issue #21, opened 2013-12-28
add StringArray.msg


k-okada:
as described in https://github.com/ros/std_msgs/pull/5 , I'd like to introduce new data type, StringArray.msg somewhere in common_msgs. Do you have any suggestion on which package to put this msg?

```
$ cat StringArray.msg
string[] strings
```

we're using this type of message, for example obtain speech recognition result candidates.

https://svn.code.sf.net/p/jsk-ros-pkg/code/trunk/jsk_smart_apps/jsk_gui_msgs/msg/VoiceMessage.msg

furushchev:
Is there any proceeding?
Is it no problem to put into an existing package or make a new package in common_msgs?

tfoote:
Looking at this message I do not see it adding significant value by declaring such a generic datatype without semantic meaning. And without specific meaning there's no point in creating a single line msg file. 

A well defined message should stand alone and be interpretable without any other context. For example a LaserScan captures the metadata as well as payload and any program knows what to expect when receiving the message. 

A such you could propose setting up voice specific messages which would have meaning for voice recognition. However they should have strong semantic meaning so that the data can be interpreted effectively. 

i'm not sure of your exact use cases but I would suggest something like:

```
VoiceRecognitionCandidates.msg
# This message represents the candidate phrases detected by voice recognition
String [] canidiates
```

or a more complex datatype

Where it's always known that VoiceRecognitionCandidates are differentiated from 

```
Letter.msg
String [] sentances
```

Which is simply a collection of sentences. These could both be StringArray but have very different values.  

Also when embedding a StringArray would require you to address the subfield using 

my_msgs.strings.strings[0]

vs if you use the String [] type it could simply be 

my_msg.strings[0]

maxbader:
Hi

I just proposed a generic_msgs pkg (https://github.com/ros/common_msgs/issues/49) maybe this would also solve your problem.

Greetings

tfoote:
Closing as proposed messages don't have semantic meaning. See parallel discussion in #49.

PR #32, opened 2014-04-07
Add speech recognition msg

furushchev:
I'd like to add SpeechRecognitionCandidates.msg to common_msgs, which includes two values:

```
string[] transcript   # candidate words of speech-to-text API
float32[] confidence # confidence of transcript
```

These are named after [W3C Web Speech API Specification](https://dvcs.w3.org/hg/speech-api/raw-file/9a0075d25326/speechapi.html) which is supported by Chrome.

This time I included this message in package sensor_msgs,
but, I cannot decide whether package to include this message in.
Now that, I think, more feature which is useful to robot using smart phone/tablet can appears such as  robot  app launcher, speech recognition, robot navigation controller apps.
Is it correct to include messages for such a "gui app" to sensor_msgs?
Should I make new package which is called something like "gui_msgs"?

### DIFF
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -36,6 +36,7 @@ add_message_files(
   Range.msg
   RegionOfInterest.msg
   RelativeHumidity.msg
+  SpeechRecognitionCandidates.msg
   Temperature.msg
   TimeReference.msg)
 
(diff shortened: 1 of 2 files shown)


tfoote:
@furushchev Thanks for the submission. This looks like a good suggested set of messages and leveraging the W3C work is good. I suggest that you create and release a package with these messages, possibly something like voice_recognition_msgs. Test and evaluate the messages with the early adopters. Once the message types are stabilized the whole package can be included into the  common_msgs metapackage definition. 

I've added guidelines for submission to common_msgs on the wiki page: http://wiki.ros.org/common_msgs

Related to #21

furushchev:
@tfoote Thanks for nice suggestion. OK.

PR #33, opened 2014-04-09
Finally add STALE

trainman419:
yep; it's going to break everything. We don't guarantee cross-version compatibility anyway, and this has been around long enough that it needs to get added to the message definition sooner or later.

### DIFF
--- a/diagnostic_msgs/msg/DiagnosticStatus.msg
+++ b/diagnostic_msgs/msg/DiagnosticStatus.msg
@@ -5,6 +5,7 @@
 byte OK=0
 byte WARN=1
 byte ERROR=2
+byte STALE=3
 
 byte level # level of operation enumerated above 
 string name # a description of the test/component reporting


chadrockey:
+1 this is long overdue and shouldn't be too painful.

Can you add a migration rule to diagnostic_msgs in case someone has these bagged?

tfoote:
+1 with a bag migration rule

trainman419:
I tried to run the rule migration script and got this:

```
$ rosrun rosbag makerule.py DiagnosticStatus.saved DiagnosticStatusIndigo.bmr
The following migrations need to occur:
 * From: diagnostic_msgs/DiagnosticStatus [67d15a62edb26e9d52b0f0efa3ef9da7]
   To:   diagnostic_msgs/DiagnosticStatus [d0ce08bc6e5ba34c7754f563a9cabaf1]

No additional rule files needed to be generated.  DiagnosticStatusIndigo.bmr not created.
```

Did I miss an argument, or it it smart enough to figure out the migration without a rule?

chadrockey:
Its smart enough to match members one to one, and since package name, message name, and all members match, it will automatically migrate what's in a bag to what's in the ROS package path.  The enum constants are just for developers and as integers, can't be migrated.

I think this is +1 to merge from me at this point.

trainman419:
@tfoote any thoughts on this? Is there a way I can generate a migration rule, or can we move forward without one?

tfoote:
I believe @chadrockey is right so we can merge this.

mikeferguson:
I've added a note on the Indigo migration page about this, since it will most definitely break your robot_monitor or dashboard if you aren't paying close attention to which distro is running on each end (and there is no error/warning on the monitor/dashboard -- only in the robot logs)

PR #41, opened 2014-07-10
Is it possible to add some more basic geometry msgs?

maxbader:
Hi

I added some basic geometriy messages like Lines.
Is it possible to get the stuff merged into to main branch?

Greetings
Markus

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -6,13 +6,26 @@ find_package(catkin REQUIRED COMPONENTS message_generation std_msgs)
 add_message_files(
   DIRECTORY msg
   FILES
+  Line.msg
+  Line2D.msg
+  Line2DArray.msg
+  Line2DStamped.msg
+  LineArray.msg
+  LineStamped.msg
   Point.msg
+  Point2D.msg
+  Point2DArray.msg
+  Point2DStamped.msg
   Point32.msg
   PointStamped.msg
   Polygon.msg
   PolygonStamped.msg
   Pose2D.msg
   Pose.msg
+  Pose2D.msg
+  Pose2DArray.msg
+  Pose2DWithCovariance.msg
+  Pose2DWithCovarianceStamped.msg
   PoseArray.msg
   PoseStamped.msg
   PoseWithCovariance.msg
--- a/geometry_msgs/msg/Line.msg
+++ b/geometry_msgs/msg/Line.msg
@@ -1,0 +1,2 @@
+# This contains the start and end point of a line in free space
+Point[2] p
--- a/geometry_msgs/msg/Line2D.msg
+++ b/geometry_msgs/msg/Line2D.msg
@@ -1,0 +1,2 @@
+# This contains the start and end point of a line in 2d space
+Point2D[2] p
--- a/geometry_msgs/msg/Line2DArray.msg
+++ b/geometry_msgs/msg/Line2DArray.msg
@@ -1,0 +1,3 @@
+# This contains the start and end point of a line in 2d space
+Header header
+Line2D[] lines
--- a/geometry_msgs/msg/Line2DStamped.msg
+++ b/geometry_msgs/msg/Line2DStamped.msg
@@ -1,0 +1,3 @@
+# This represents a 2d line with reference coordinate frame and timestamp
+Header header
+Line2D line
--- a/geometry_msgs/msg/LineArray.msg
+++ b/geometry_msgs/msg/LineArray.msg
@@ -1,0 +1,3 @@
+# An array of lines with a header for global reference
+Header header
+Line[] lines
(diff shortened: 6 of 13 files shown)


tfoote:
To merge things into common_msgs our policy is outlined on the wiki page here: http://wiki.ros.org/common_msgs

In general we want new messages to be reviewed and have been demonstrated as useful before merging them into the core. We want common_msgs to actually represent demonstrated usefulness and not try to set standards before testing them throughly. Thus we usually ask for 

If this was an API review here are some quick comments I would make. 

For the 2D, we've avoided collapsing from 3D to 2D in general since the savings in footprint and transport overhead is relatively small and you offset that by needing twice the API to manipulate all the datatypes. Similarly a TypeA and TypeA2D would be incompatible and there would also need to be converters too and from the datatypes. We originally had Pose2D as we thought that it would be valuable for things like 2D navigation. However we determined that use the Pose datatype is much more scaleable since both 2D and 3D navigation can have the same inputs and outputs. And the application can use appropriate logic for collapsing into 2D or rejecting invalid inputs as is appropriate for the situation. 

When supporting 2D datatypes also there are often manifolds and semantically important distinctions between the datatypes. For example the most common 2D datatypes is the position on the surface of the earth. For which a generic 2D datatype cannot represent, so it makes sense to have semantically specific datatypes for the specific type of manifold onto which the point is being mapped. 

For the generic Line datatype it is so simple that it doesn't make sense to define the specific datatype unless it has significant semantic meaning.  Point [2] vs Line saves a few characters, but adds another dattype needed to be supported by all potential users. An array of points can be transformed by iterating over the points. And is generic enough to trivially extend to a polyline supporting more than one segment.  However if there is a specific application or use case which needs specific functionality then it would make sense to make a datatype specific for that application/domain.

jack-oquin:
I agree with all the comments @tfoote made. 

-1 to merging this proposal

maxbader:
Hi,
Sorry for being a little bit to enthusiast. 
I am happy if you like to stick only to 3D objects. 
But I think the Line.msg, LineArray.msg and PointArray.msg are important missing messages. Just think of a line based EKF ... or many computer Vision outputs where a PointCloud is to much.
Having a generic vector type of float64 would also solve many problems e.g.

Header header
float64 v[]

I don't want to think of the hours I spend telling students how to write over and over again a msgs and on the end the programs where not able to communicate because everybody produced the msg content in a different way.
ROS is about collaboration and messages our spoken words, so how do you want to achieve a collaboration if everybody has to invent over and over again words.

What about a pkg unsuppored_msgs or beta_msgs, what ever.

Greetings
Markus

tfoote:
in many of these cases it would be appropriate to define a message package which contains the agreed upon datatypes. The interface between modules can be defined by their topics and their types. The goal is to have datatypes which stand alone and when received can be processed correctly by any subscriber.  And the topics allow you to give semantic meaning to the data. 

For your example a Vector.msg with float64 v[] as it's content is super generic and could be used for almost any purpose. It could be the coefficients of a polynomial function, or they could be the feature strength for each pixel in an image.  I just randomly made up these two cases, but if they are both published as Vector datatypes you can accidentally connect the output of a line fitting module into an image classifier giving very unexpected results. 

If instead you had simply created two messages. PolynomailCoefficients.msg and PixelFeatureStrength.msg with the same definition as Vector.msg you would not make the same mistake. 

It is recommended to take the time to define the messages by which your projects components will be interacting up front.  For a course, if the design of the system is up to the students they should think about their datatypes early, if the students are to interact with some infrastructure layed out for them, giving them a projectx_msgs package to define their expected inputs and outputs might be a good idea. 

When designing messages it is rare that a single data field can form a self describing message that contains all the information relevant for parsing. A value as simple as [Illuminance](http://docs.ros.org/api/sensor_msgs/html/msg/Illuminance.html) or [FluidPressure](http://docs.ros.org/api/sensor_msgs/html/msg/FluidPressure.html) sensor readings have more than one field. For example it's important to know when the reading was taken and in what frame it was observed. 

And reusing a semi-generic datatype increases complexity when composing datatypes. For example it requires an extra layer of depth for accessing data when composing messages.  my_msg1.vector.v[5] is much less intuitive than my_msg2.vector[5] where my_msg1 uses the predefined Vector.msg vs simply using float64[] in my_msg2

All of this is generally thinking about how to leverage a strongly typed system. The strong typing can be circumvented by defining generic messages and then packing different data inside the message. you can do everything inside of a uint8[] field if you wanted to do your own form of serialization into that format. The power of being strongly typed is that introspection, debugging and checking by the system.  The disadvantage is that it requires you to define the types you want to use ahead of time. And although that can seem to be a burden often it is a great enabler for collaborative work by having clearly defined interfaces. 

If you have some specific uses I'd be happy to discuss some application specific messages but as stated on the wiki page our typical approach for promoting messages is to release a standalone message package and after it's become commonly used promote it into common_msgs. 

If you feel your students are reinventing message datatypes specific to any domain i suggest that you or they put together a package of messages specific to that domain so that others can consider adopting them. Adding layers of primative elements however will not prevent the need for these domain specific messages.

PR #42, opened 2014-07-25
geometry_msgs: Adding inertia message

jbohren:


jbohren:
Sorry, the correct version of these is actually somewhere else.

PR #43, opened 2014-08-07
actionlib_msgs: use the DIRECTORY argument instead of the BASE_DIR argum...

dreuter:
...ent of add_message_files. The DIRECTORY argument of add_action_files now also takes absolute paths.

This hopefully makes the BASE_DIR argument of add_message_files useless. (See https://github.com/ros/genmsg/pull/51).

Sorry that it took so long, I wasn't able to find some time to do this earlier.

Does anyone have a good idea how deactivate packages like sensor_msgs, when installed via dkpg? I would need this feature to test my pull request, since I don't really want to set up a virtual box and compile from source (at least not today).

### DIFF
--- a/actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
+++ b/actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
@@ -21,8 +21,15 @@ macro(add_action_files)
     set(ARG_DIRECTORY "action")
   endif()
 
-  if(NOT IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY})
-    message(FATAL_ERROR "add_action_files() directory not found: ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY}")
+  if(IS_ABSOLUTE ${ARG_DIRECTORY})
+    set(ACTION_DIR ${ARG_DIRECTORY})
+  else()
+    # Prepend CMAKE_CURRENT_SOURCE_DIR if path is not absolute
+    set(ACTION_DIR ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY})
+  endif()
+
+  if(NOT IS_DIRECTORY ${ACTION_DIR})
+    message(FATAL_ERROR "add_action_files() directory not found: ${ACTION_DIR}")
   endif()
 
   # if FILES are not passed search action files in the given directory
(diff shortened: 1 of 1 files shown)


dreuter:
Found a (very hacky) way to test it: I copied the files into /opt/ros/hydro/share/... and replaced the templated variables...

But it works :)

actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
```diff
@@ -73,7 +80,7 @@ macro(add_action_files)
     endif()
 
     add_message_files(
-      BASE_DIR ${MESSAGE_DIR}
+      DIRECTORY ${MESSAGE_DIR}
```
dirk-thomas, line 83: Since this change requires a modified version of genmsg the patch must contain a version dependency on genmsg with the next upcoming patch release.

dirk-thomas, line 83: The `genmsg` version supporting this is `0.5.4`.

actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
```diff
@@ -21,8 +21,15 @@ macro(add_action_files)
...
   endif()
 
-  if(NOT IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY})
-    message(FATAL_ERROR "add_action_files() directory not found: ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY}")
+  if(IS_ABSOLUTE ${ARG_DIRECTORY})
+    set(ACTION_DIR ${ARG_DIRECTORY})
+  else()
+    # Prepend CMAKE_CURRENT_SOURCE_DIR if path is not absolute
+    set(ACTION_DIR ${CMAKE_CURRENT_SOURCE_DIR}/${ARG_DIRECTORY})
+  endif()
```
dirk-thomas, line 29: This code block should be modified to look like this:

```
set(ACTION_DIR "${ARG_DIRECTORY}")
if(NOT IS_ABSOLUTE "${ACTION_DIR}")
  set(ACTION_DIR "${CMAKE_CURRENT_SOURCE_DIR}/${ACTION_DIR}")
endif()
```

actionlib_msgs/cmake/actionlib_msgs-extras.cmake.em
```diff
@@ -41,7 +48,7 @@ macro(add_action_files)
   endforeach()
 
   if(NOT ARG_NOINSTALL)
-    install(FILES ${FILES_W_PATH} DESTINATION share/${PROJECT_NAME}/${ARG_DIRECTORY})
+    install(FILES ${FILES_W_PATH} DESTINATION share/${PROJECT_NAME}/${ARG_DIRECTORY}) # Shouldn't this be /action?
```
dirk-thomas, line 51: If the user chooses a custom location the files will be installed to a location with the same custom name. Same as in `genmsg`: https://github.com/ros/genmsg/blob/807a8590d0a855fb091e8bc0f2bc100a3579a395/cmake/genmsg-extras.cmake.em#L130

tfoote:
replaced by #96

PR #44, opened 2014-10-03
geometry_msgs: Adding inertia message

jbohren:
What's the best way to get this reviewed?

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -6,6 +6,8 @@ find_package(catkin REQUIRED COMPONENTS message_generation std_msgs)
 add_message_files(
   DIRECTORY msg
   FILES
+  Inertia.msg
+  InertiaStamped.msg
   Point.msg
   Point32.msg
   PointStamped.msg
--- a/geometry_msgs/msg/Inertia.msg
+++ b/geometry_msgs/msg/Inertia.msg
@@ -1,0 +1,14 @@
+# Mass [kg]
+float64 m
+# First mass moment [kg-m]
+geometry_msgs/Vector3 h
+# Inertia Tensor [kg-m^2]
+#     | ixx ixy ixz |
+# I = | ixy iyy iyz |
+#     | ixz iyz izz |
+float64 ixx
+float64 ixy
+float64 ixz
+float64 iyy
+float64 iyz
+float64 izz
--- a/geometry_msgs/msg/InertiaStamped.msg
+++ b/geometry_msgs/msg/InertiaStamped.msg
@@ -1,0 +1,2 @@
+Header header
+geometry_msgs/Inertia inertia


jack-oquin:
I recommend opening a review discussion on the ros-sig-drivers mailing list.

aleeper:
It makes sense to have mass / mass distribution messages, though I was a bit surprised to see mass and first-moment in the message. "Inertia" is such a loaded term...

Can you describe a bit how you've been using this message? What are the use cases for transmitting this stuff dynamically vs. reading things out of an URDF? Also, I'm curious what uses you have for the first moment, h?

jbohren:
I thought I responded to this a mouth ago, sorry for the delay.

I've been using this for broadcasting the inertial properties of known objects for the feed-forward term in a model-based manipulator controller. I've also used it for brocasting the lump inertial properties of a flexible, but instrumented object attached to the end-effector. 

Using this message enables the controller to not need to know anything about the structure it's manipulating except for the dynamic properties.

The first moment is in there because it gives the position of the center of mass of the object. Using the first moment instead of the vector to the poison of the center of mass just seemed more elegant. I.e. the message contains the 0th, 1st, and 2nd mass moments.

tfoote:
I think a good approach would be to discuss it on ros-sig-drivers with an announcement to ros-users

jbohren:
Yeah, I posted it to the drivers sig with a link to this PR after @jack-oquin suggested it. @aleeper was the only one to bite.

chadrockey:
I'll comment a little bit here.  The inertia itself looks good.  How do you propose you use this in ROS?

frame_ids aren't necessarily locations of inertias, they are most often links or interfaces.

Do you have /my_node/object_a_inertia, /my_node/object_b_inertia topics currently?

My one issue is that if you gave me just one of these messages, it would be pretty meaningless.  With LaserScan, I have a measurement that I could visualize at a location and time.

With InertiaStamped, I have... mass properties at a location and time?

What I feel is missing is some association between an object or and id with an inertia.  Should inertia  include an 'object_name'?  Should Inertia stamped include that instead so that some messages could include multiple inertias on a better defined object components?  Without knowing the object, I don't think this message quite stands alone.

Are these typically published individually or in sets?  You may want to consider an Array version if there are always multiples.

jbohren:
> I'll comment a little bit here. The inertia itself looks good. How do you propose you use this in ROS?

To strongly describe inertial information. Currently the only way to send this information is as a combination of a weakly typed 4-vector and floats for the non-redundant entries in the inertia tensor.

> frame_ids aren't necessarily locations of inertias, they are most often links or interfaces.

I don't understand what you mean, any spatial information needs to be associated with some coordinate system, otherwise it's meaningless.

> Do you have /my_node/object_a_inertia, /my_node/object_b_inertia topics currently?

In the control use case, there can be a topic such as `/controller/ee_inertia` to which a node which has additional inertial information can publish. This decouples the responsibility of knowing what additional inertia needs to be accounted for (object sensing / model) from the responsibility of knowing how to augment the command to the manipulator (control).

For example, if you have an object detector / tracker which reports the pose of the object or structure being grasped, and another node which can compute the inertial properties of that structure,  it can send just the inertial information to the controller to compensate for the added mass.

> My one issue is that if you gave me just one of these messages, it would be pretty meaningless.

I don't see why one of these messages would be meaningless. It would have just as much meaning as any other message in the geometry_msgs package describing physical-spatial information.

> With LaserScan, I have a measurement that I could visualize at a location and time.
> 
> With InertiaStamped, I have... mass properties at a location and time?

Yes, that's exactly what you have. Furthermore it's easy to visualize such a message: e.g. you have a point at the center of mass, scaled by the size of the mass, and an ellipsoid showing the mass distribution in three dimensions.

> What I feel is missing is some association between an object or and id with an inertia. Should inertia include an 'object_name'?

No. This is higher-level information. It should not be expressed at this level for the same reason we don't name Twists.

> Should Inertia stamped include that instead so that some messages could include multiple inertias on a better defined object components? Without knowing the object, I don't think this message quite stands alone.

This is just a message to strongly type messages containing inertial information. It can be used alone if your topic semantics are well defined or if you want to embed it into another higher-level message the same way other messages in `geometry_msgs` are used.

> Are these typically published individually or in sets? You may want to consider an Array version if there are always multiples.

Yes, I think an `InertiaArray` message could be an appropriate addition to this PR.

aleeper:
@jbohren After reading through your response to me and to the other comments, I think this message is good to go.

I am ambivalent about `InertiaArray`. Jon, it seems you did not have a use case for a bunch of inertias transmitted together? Though, I suppose it couldn't hurt to have to reduce topic bloat.

jbohren:
> I am ambivalent about InertiaArray. Jon, it seems you did not have a use case for a bunch of inertias transmitted together? Though, I suppose it couldn't hurt to have to reduce topic bloat.

Normally if I'm considering a collection of inertias, I still need to sum them together in order to get something useful. To that end, having an array could be convenient, but probably means you've split up the computation into too many nodes. I think we're fine without it.

jbohren:
> @jbohren After reading through your response to me and to the other comments, I think this message is good to go.

And the name is ok? Other options include things like:
- `RigidBodyInertia` (like KDL)
- `Inertial` (to emphasize that it's more than just the inertia tensor)

aleeper:
@jbohren Despite my initial confusion, I do think `Inertia` is correct. Just because "inertia" is frequently used to refer to 2nd moment and products of inertia doesn't mean that mass and first moment are not also inertia :)

Also, just to make sure it doesn't gain traction, I am actively against `RigidBodyInertia` because it is semantically limiting. It is perfectly valid to talk about the inertia of a system of bodies or particles, and for control purposes it may be totally valid to use the instantaneous inertia properties of a kinematic sub-system.

jbohren:
@jbohren Despite my initial confusion, I do think Inertia is correct. Just because "inertia" is frequently used to refer to 2nd moment and products of inertia doesn't mean that mass and first moment are not also inertia :)

Cool.

> Also, just to make sure it doesn't gain traction, I am actively against RigidBodyInertia because it is semantically limiting. It is perfectly valid to talk about the inertia of a system of bodies or particles, and for control purposes it may be totally valid to use the instantaneous inertia properties of a kinematic sub-system.

That's reasonable.

jbohren:
Also we should probably specify that the second moment is computed about the center of mass (this is the same convention used by URDF and SDF).

aleeper:
So, actually this brings up a good point that I overlooked, and it's really important to have clear.

In the stamped case, I would hope that the first moment is measured from the origin of `frame_id`; the inertia tensor is about the origin of `frame_id` **or** about the center of mass point deduced from `first moment / mass`; and the XYZ basis in which both are expressed is the same as `frame_id`.

If you have just a plain inertia message:
- there is (or, there needs to be) the implicit assumption that the first moment and the inertia tensor are "expressed in" the same XYZ unit vector directions.
- what point is the first moment vector measured **from**? There is sort of an implied frame origin there.
- what is the "about point" for the inertia tensor entries? I think the only two options, like in the stamped case, are the (implicit) frame origin (0, 0, 0) or the center of mass deduced from `first moment / mass`.

jbohren:
> In the stamped case, I would hope that the first moment is measured from the origin of frame_id; the inertia tensor is about the origin of frame_id **or** about the center of mass point deduced from first moment / mass; and the XYZ basis in which both are expressed is the same as frame_id.

Right, and since neither convention has any particular benefit, I think it makes sense to use the same convention used in URDF, where the second moment is computed about the center of mass (`h/m`).

> If you have just a plain inertia message:
> - there is (or, there needs to be) the implicit assumption that the first moment and the inertia tensor are "expressed in" the same XYZ unit vector directions.

I think that's a correct assumption.

> - what point is the first moment vector measured from? There is sort of an implied frame origin there.

Yeah, this is the same convention as with `Twist` and `Wrench` messages: the frame of reference will be based on whatever context the message is given in. If it's given in an `InertiaStamped` message, then the frame is given there, if it's given in the context of some other application or message type, then it will be based on that context.

> - what is the "about point" for the inertia tensor entries? I think the only two options, like in the stamped case, are the (implicit) frame origin (0, 0, 0) or the center of mass deduced from first moment / mass.

For the same reason as above in the stamped case, I think it should be about the center of mass.

chadrockey:
OK, thanks for replying @jbohren.

Can you add small comments at the top of each message giving a one sentence explanation of what the messages represent.

Also, can you add comments for the Stamped version clarifying what the stamp and frame_id represent.

Stamp cannot be left blank (ros::Time(0)) right?  If it is, then it falls back that the unstamped version should be used.  So the stamped version represents either measured or changing inertia, or inertia at a location and time, right?

And frame_id is the location of the center of mass at the timestamp?

jbohren:
> Can you add small comments at the top of each message giving a one sentence explanation of what the messages represent.

Will do.

> Also, can you add comments for the Stamped version clarifying what the stamp and frame_id represent.

Sure.

> Stamp cannot be left blank (ros::Time(0)) right? If it is, then it falls back that the unstamped version should be used. So the stamped version represents either measured or changing inertia, or inertia at a location and time, right?

For a static inertia, the `stamp` can be ignored.

> And frame_id is the location of the center of mass at the timestamp?

`frame_id` is not the location of the center of mass, it names the coordinate frame in which the center of mass (`h/m`) and inertia tensor (`I`) are defined.

aleeper:
@jbohren Ok we are in agreement on the semantics. I agree with having the inertia values be about the center of mass ( `h/m` ). It might be good for the comments in the message to be extra clear about what stuff means, and examples of how to calculate it for those who might not be as familiar with the technicalities of shifting inertias, etc.

> For a static inertia, the stamp can be ignored.

Just to avoid confusion (and correct me if I'm wrong): the timestamp (which is of type "stamp") could be ignored, but it doesn't revert the message to the "unstamped type" because the frame_id gives information about the implied origin and the vector basis used in the Inertia message, regardless of any time info. Maybe the ros convention for "Stamped" messages should have been called "Headered" instead ;).

jbohren:
> It might be good for the comments in the message to be extra clear about what stuff means, and examples of how to calculate it for those who might not be as familiar with the technicalities of shifting inertias, etc.

Ok, I'll add in some comments to the PR.

aleeper:
Did the discussion for this end up moving somewhere? Or, what's the latest news? :)

jbohren:
After using this for a while, I think it's better to just be mass / center-of-mass / inertia tensor instead of having the first mass moment in there. See #53 for a new PR (and referenced for each distro).

PR #45, opened 2014-10-14
replaced a confusing sentence about REP 117

mehditlili:


chadrockey:
This comment refers specifically to binary range sensors.  They are more like energy based bumpers than anything else.  The measurement only toggles between -Inf and +Inf, so they are either "detected" or not, thus the original wording.  Too close implies that there is a measurement possible, but for these specific sensors, that is not the case.  Is there another wording that would clarify this usage?

mehditlili:
Thanks for the clarification. Using REP 117 confused me a little bit. I will close this request.

PR #48, opened 2014-11-04
Fix compilation with Clang

jmtatsch:


vrabaud:
+1. Leaving it open until somebdoy can release it. Sorry about creating the problem.

tfoote:
Thanks for catching that @jmtatsch

PR #54, opened 2015-03-09
geometry_msgs: adding inertia messages

jbohren:
Rebase of #53 into hydro


tfoote:
Thanks for the separate pull-requests. With hydro approaching EOL i'd prefer not to be changing it significantly. Do you have a need for this in Hydro? Otherwise I'd prefer to close this without merging.

PR #56, opened 2015-03-19
Adds a SetMap service message to support swap maps functionality in amcl

stwirth:
This is a backport of 04cceab to indigo.

@liz-murphy @mikeferguson @tfoote I just noticed that the `initial_pose` should rather be a `PoseWithCovarianceStamped` to support passing the covariance, analogue to amcl's topic `/initial_pose`. What do you think? The idea behind this new service was to get the same as publishing a new map and a new initial pose but with the feedback from the handler of those messages to make sure they actually arrived and are in use.

### DIFF
--- a/nav_msgs/CMakeLists.txt
+++ b/nav_msgs/CMakeLists.txt
@@ -16,7 +16,8 @@ add_service_files(
   DIRECTORY srv
   FILES
   GetMap.srv
-  GetPlan.srv)
+  GetPlan.srv
+  SetMap.srv)
 
 add_action_files(
   FILES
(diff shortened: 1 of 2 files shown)


tfoote:
It makes sense to me to have covariance like the `/initial_pose` topic.

liz-murphy:
Yes I agree. Forcing it to have a default covariance was a bit shortsighted.

mikeferguson:
Concur as well @stwirth please also open a PR to update Jade

stwirth:
Changed `initial_pose` to be `PoseWithCovarianceStamped`.

stwirth:
Same for jade: https://github.com/ros/common_msgs/pull/57

tfoote:
Thanks.

PR #57, opened 2015-03-20
change type of initial_pose in SetMap service to PoseWithCovarianceStamped

stwirth:


mikeferguson:
@tfoote can we get new indigo and jade releases so that I can merge the navigation pull requests?

tfoote:
@mikeferguson released

PR #61, opened 2015-04-06
remove warnings

vrabaud:
and cleanup doc a bit.

Warnings were mentioned by @mikeferguson in https://github.com/ros/common_msgs/pull/16#issuecomment-89513070


tfoote:
Thanks for the cleanup!

PR #64, opened 2015-04-16
geometry_msgs: entry in CMakeLists.txt for inertia messages

gaya-:


airballking:
@gaya- Could you please make another pull request like this targeting `indigo-devel`? That branch has the same problem...

gaya-:
Oh, I though Tully would pull it over. Sure.

PR #65, opened 2015-04-17
geometry_msgs: CMakeLists entry for inertia msgs.

gaya-:


tfoote:
Thanks for the parallel PR.

PR #66, opened 2015-05-10
[sensor_msgs] Add utility functions for  inspecting image encodings

wkentaro:


tfoote:
Generally I'd prefer to avoid one line inline functions which the user can trivially reproduce. I'm not sure these methods will save the end user code complexity, since to be able to iterate you will need to know the size, so even if you use these methods you will need another conditional to operate on the data. 

What are your use cases where you want to know the order but not the size of the encoding? 

Exposing more semantically meaningful methods like isBGROrder which tests all 4 bgr encodings would seem more powerful.

There's also an hasAlpha method which could be used with the ordering to give the same results as the methods defined.

wkentaro:
I should have written about the use cases.

I added these functions to use them when converting subscribed image msg to
image which has desired encoding.
If to mono8, cv_bridge automatically converts msg, but if not auto
converting does not work.
So in these cases, I need to convert image encoding using other methods
like cv::cvtColor, and I need to know the image msg is BGR/RGB/BGRA/RGBA at
this time.
I thought these utility functions are useful for that. Are there any other
way?

2015年5月11日月曜日、Tully Footenotifications@github.comさんは書きました:

vrabaud:
I would agree with not putting it: the functions are more about the semantic of the colors: does it have colors, alpha, is it bayer-based. 
@wkentaro , your functions are indeed trivial to reproduce. Now, there are two functions in cv_bridge that could be of interest to you and that should probably be made public:
https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L49
https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L103
Could you use those to solve your issue ?

wkentaro:
Thanks.
I'd like to use getFormat function in cv_bridge, but it's not written in
the header file. And it does not seem good idea to write getFormat, so I
think I will write on my own.

2015年5月12日火曜日、Vincent Rabaudnotifications@github.comさんは書きました:

> I would agree with not putting it: the functions are more about the
> semantic of the colors: does it have colors, alpha, is it bayer-based.
> Your functions are indeed trivial to reproduce. Now, there are two
> functions in cv_bridge that could be of interest to you and that should
> probably be made public:
> 
> https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L49
> 
> https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L103
> Could you use those to solve your issue ?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ros/common_msgs/pull/66#issuecomment-101038735.

## 

和田 健太郎 / Kentaro Wada
http://wkentaro.com http://www.wkentaro.com

wkentaro:
OR Add similar function to image_encodings.h?

2015年5月12日火曜日、和田 健太郎 / Kentaro Wadawww.kentaro.wada@gmail.comさんは書きました:

> Thanks.
> I'd like to use getFormat function in cv_bridge, but it's not written in
> the header file. And it does not seem good idea to write getFormat, so I
> think I will write on my own.
> 
> 2015年5月12日火曜日、Vincent Rabaud<notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>さんは書きました:
> 
> > I would agree with not putting it: the functions are more about the
> > semantic of the colors: does it have colors, alpha, is it bayer-based.
> > Your functions are indeed trivial to reproduce. Now, there are two
> > functions in cv_bridge that could be of interest to you and that should
> > probably be made public:
> > 
> > https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L49
> > 
> > https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/cv_bridge.cpp#L103
> > Could you use those to solve your issue ?
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > https://github.com/ros/common_msgs/pull/66#issuecomment-101038735.
> 
> ## 
> 
> 和田 健太郎 / Kentaro Wada
> http://wkentaro.com http://www.wkentaro.com

## 

和田 健太郎 / Kentaro Wada
http://wkentaro.com http://www.wkentaro.com

wkentaro:
Close this for now, because this function can be easily reproduced.

PR #70, opened 2015-05-17
sensor_msgs: range: added LASER radiation type

TSC21:
Related to https://github.com/mavlink/mavros/pull/292#issuecomment-101754507
[Groovy release]


tfoote:
Groovy is EOL

PR #73, opened 2015-05-17
sensor_msgs: range: added LASER radiation type

TSC21:
Related to https://github.com/mavlink/mavros/pull/292#issuecomment-101754507.
[Jade release]

### DIFF
--- a/sensor_msgs/msg/Range.msg
+++ b/sensor_msgs/msg/Range.msg
@@ -15,6 +15,7 @@ Header header           # timestamp in the header is the time the ranger
 # If you want a value added to this list, send an email to the ros-users list
 uint8 ULTRASOUND=0
 uint8 INFRARED=1
+uint8 LASER=2
 
 uint8 radiation_type    # the type of radiation used by the sensor
                         # (sound, IR, etc) [enum]


tfoote:
LASER is not a radiation type. So it's not appropriate to add to this enumeration. 

The reason for this field is to know what types of objects will reflect vs not reflect. Ultrasonic vs infrared have very different behaviour when looking at glass for example. 

As mentioned, in that comment, lasers can have different types of radiation. We could consider adding VISIBLE and ULTRAVIOLET to complement the existing INFRARED definition. 

However I am not aware of a "light" based distance measuring device operating in the visible or ultraviolet frequencies. 

There are also definitely microwave based sensors which might be applicable. However we try not to speculatively update messages until there is a clearly known use case. 

From what sensor or sensors are you trying to represent the data using this message format?

TSC21:
> LASER is not a radiation type. So it's not appropriate to add to this enumeration.

Laser - "A laser is a device that emits light through a process of optical amplification based on the stimulated emission of electromagnetic radiation." I accept that radiation definition, but then it should also add NEAR_INFRARED. In this case we also have two kinds of radiation: electromagnetic and mechanical.

> As mentioned, in that comment, lasers can have different types of radiation. We could consider adding VISIBLE and ULTRAVIOLET to complement the existing INFRARED definition.

The comment was referring to Lidar, which is a specific application for lasers.

> However I am not aware of a "light" based distance measuring device operating in the visible or ultraviolet frequencies.

The reference to the other types of wavelength/radiotion used by Lidars comes from http://en.wikipedia.org/wiki/Lidar. Type depends on the application/range.

> From what sensor or sensors are you trying to represent the data using this message format?
- https://www.parallax.com/sites/default/files/downloads/28043-SF02-Laser-Rangefinder-Manual.pdf (operates at 850nm, which is near-IR wavelength - 780–1400nm)
- http://www.robotshop.com/media/files/pdf/operating-manual-llm20c132i500s011.pdf (operates at 905nm, which is near-IR wavelength)

tfoote:
For both of those just use INFRARED we're not trying to represent the exact wavelength or specific taxonomy. We can use the general definition of INFRARED to cover things between visible and microwave. 

https://en.wikipedia.org/wiki/Infrared
http://www2.lbl.gov/MicroWorlds/ALSTool/EMSpec/EMSpec2.html

TSC21:
Will accept then @tfoote. Thanks for your time! Offtopic:  is there a possibility of inserting covariance in the message definition?

PR #74, opened 2015-07-08
Add a battery state message

tfoote:
This has come from the ROS SIG Mav reviews looking for a smart battery message. https://groups.google.com/forum/#!topic/ros-sig-mav/f12m3mnqYl4

This is an iteration on [linux_hardware/LaptopChargeStatus.msg](http://docs.ros.org/hydro/api/linux_hardware/html/msg/LaptopChargeStatus.html) 

There's another similar message here: https://github.com/ros-drivers/smart_battery_msgs/blob/master/msg/SmartBatteryStatus.msg

This assumes a smart battery. Smart batteries are becoming ubiquitous in next generations UAVs and we expect many more to become available in other domains as well. 

To support non-smart batteries the only mandatory field is the voltage. All other fields have documented undefined values. 

We added support for multi-cell information as well as a string for location and serial numbers. 

@mikeferguson Would you consider using this for here: https://github.com/fetchrobotics/power_msgs/issues/3 The one thing I don't see here is the time remaining estimate.

### DIFF
--- a/sensor_msgs/msg/BatteryState.msg
+++ b/sensor_msgs/msg/BatteryState.msg
@@ -1,0 +1,49 @@
+
+# Constants are chosen to match the enums in the linux kernel
+# defined in include/linux/power_supply.h as of version 3.7
+# The one difference is for style reasons the constants are
+# all uppercase not mixed case.
+
+# Power supply status constants
+uint8 POWER_SUPPLY_STATUS_UNKNOWN = 0
+uint8 POWER_SUPPLY_STATUS_CHARGING = 1
+uint8 POWER_SUPPLY_STATUS_DISCHARGING = 2
+uint8 POWER_SUPPLY_STATUS_NOT_CHARGING = 3
+uint8 POWER_SUPPLY_STATUS_FULL = 4
+
+# Power supply health constants
+uint8 POWER_SUPPLY_HEALTH_UNKNOWN = 0
+uint8 POWER_SUPPLY_HEALTH_GOOD = 1
+uint8 POWER_SUPPLY_HEALTH_OVERHEAT = 2
+uint8 POWER_SUPPLY_HEALTH_DEAD = 3
+uint8 POWER_SUPPLY_HEALTH_OVERVOLTAGE = 4
+uint8 POWER_SUPPLY_HEALTH_UNSPEC_FAILURE = 5
+uint8 POWER_SUPPLY_HEALTH_COLD = 6
+uint8 POWER_SUPPLY_HEALTH_WATCHDOG_TIMER_EXPIRE = 7
+uint8 POWER_SUPPLY_HEALTH_SAFETY_TIMER_EXPIRE = 8
+
+# Power supply technology (chemistry) constants
+uint8 POWER_SUPPLY_TECHNOLOGY_UNKNOWN = 0
+uint8 POWER_SUPPLY_TECHNOLOGY_NIMH = 1
+uint8 POWER_SUPPLY_TECHNOLOGY_LION = 2
+uint8 POWER_SUPPLY_TECHNOLOGY_LIPO = 3
+uint8 POWER_SUPPLY_TECHNOLOGY_LIFE = 4
+uint8 POWER_SUPPLY_TECHNOLOGY_NICD = 5
+uint8 POWER_SUPPLY_TECHNOLOGY_LIMN = 6
+
+Header  header
+float32 voltage          # Voltage in Volts (Mandatory)
+float32 current          # Negative when discharging (A)  (If unmeasured NaN)
+float32 charge           # Current charge in Ah  (If unmeasured NaN)
+float32 capacity         # Capacity in Ah (last full capacity)  (If unmeasured NaN)
+float32 design_capacity  # Capacity in Ah (design capacity)  (If unmeasured NaN)
+float32 percentage       # Charge percentage on 0 to 1 range  (If unmeasured NaN)
+uint8   power_supply_status     # The charging status as reported. Values defined above
+uint8   power_supply_health     # The battery health metric. Values defined above
+uint8   power_supply_technology # The battery chemistry. Values defined above
+bool    present          # True if the battery is present
+
+float32[] cell_voltage   # An array of individual cell voltages for each cell in the pack
+                         # If individual voltages unknown but number of cells known set each to NaN
+string location          # The location into which the battery is inserted. (slot number or plug)
+string serial_number     # The best approximation of the battery serial number
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -13,6 +13,7 @@ include_directories(include)
 add_message_files(
   DIRECTORY msg
   FILES
+  BatteryState.msg
   CameraInfo.msg
   ChannelFloat32.msg
   CompressedImage.msg


tfoote:
@vooon How does this look to replace http://docs.ros.org/indigo/api/mavros/html/msg/BatteryStatus.html ?

vooon:
Why default for cell_voltages is -1, not NAN?
Also maybe battery chemistry type?

But looks good.

tfoote:
Good point on the NaN; updated. For the chemistry do you have a use case where the robot would respond differently knowing it's battery chemistry?

mikeferguson:
Given the number of fields already in this message -- is there any reason not to have a time_remaining, or seconds_remaining. I would think such a value (if available from the battery/charger) would be especially of interest to aerial robots (so they might land before they fall out of the sky).

If we had a time_remaining or seconds_remaining (which could be a float32 rather than the larger duration message type), I'd be in favor of using this over the message we just put into power_msgs.

vooon:
@tfoote nope, just to be synced with https://pixhawk.ethz.ch/mavlink/#BATTERY_STATUS

@mikeferguson really it's is not easy task because consumption very depends on drone type and flight manner.

mikeferguson:
@vooon Ok, so maybe it can't be filled in for your drone -- but by putting this in sensor_msgs, we are defining a BatteryState for all of ROS here. If using a smart battery, may of them output time remaining (yes, it might be noisy), but it is a very useful number.

vooon:
@mikeferguson hmm, got it. 
But i not expect that on drone power batteries, while on notebooks it is done (not sure where it done kernel or DE).

vooon:
Also perhaps use standard Duration type for that field.

mikeferguson:
@vooon We used a duration for time_remaining in power_msgs -- I just know sometimes people worry about the overall size of a message, and so a float32 is 1/4 the size of a Duration (message mentioned is here https://github.com/fetchrobotics/power_msgs/blob/master/msg/BatteryState.msg)

vooon:
As i know problem of 2038 exists for ROS now, because it uses two 32-bit integers for time (ros::Time, ros::Duration etc.).
So float32 is only 1/2 of duration.

dirk-thomas:
The problem of the upper boundary of ros::Time is not relevant here. I hardly expect the duration of any battery to exceed int32 seconds. I think `duration` is the "right" type for the remaining time. Using float32 for size reasons would argue to define `duration` as `float32` itself which I am sure is not the case for very good reasons.

jack-oquin:
+1 LGTM

tfoote:
wrt to the time remaining, I'm wondering if the statistics are better than dividing the current in to the capacity? There are definitely ways to improve the time estimates, but most of them require more knowledge about the application and expected changes to the load. On computers they just make the assumption of constant consumption.

chadrockey:
A few comments:

Smart batteries can go into shutdown states, this should be reflected in the enum.  (OVERCURRENT?  OVERDISCHARGE?  OVERVOLTAGE?)

Optionally, since this is a wide meaning field and ENUM changes break md5 but comment changes do not, you could make charge_state and present become "string battery_state" and allow the user to fill in anything they want with suggestions for standards in the comments (BALANCING, OVERTEMP, SLEEP, DEEP SLEEP, CHARGING, NOT PRESENT).  You lose some of the programmatic checks, but you would gain a lot more human readability.

Percentage is redundant information with current charge and current capacity.  We've always run into issues in LaserScan where scan_increment, angle_min, and angle_max don't agree.

jonbinney:
I agree with @chadrockey  that percentage is redundant here; seems that putting charge and capacity should be good (and in any case, some people may prefer to compute the percentage w.r.t. the design_capacity instead of the capacity).

I also think that the time_remaining isn't well defined enough if it is coming from the battery driver. If the user just wants charge/current, that is easy to calculate. If they want something more fancy, that will probably be application specific. Maybe there could even be a standard node that listens to BatteryState messages, averages the current over some window size, and computes an estimated time remaining that gets published as a float32.

vooon:
@jonbinney I disagree, usually minimal data set coming from FCU is ony voltage, current and percentage. And absolutely minimal - voltage only.
PX4 for example has throttle compensation algo.

jonbinney:
@vooon ah I see. So there may be smart batteries (or FCU that manage batteries) which report percentage but not total capacity or charge level? In that case I don't see a good way to define a message that can always work but isn't redundant.

vooon:
@jonbinney For currently wide used bare batteries capacity defined by FCU parameter.

jonbinney:
Ok, so i guess there are three options:
1.  Have capacity, charge, and percent in the message, and set unknown ones to some default value
2.  Have only capacity and percent in the message, and if the FCU/battery monitor provides only capacity and charge, then compute percent
3.  Have only capacity and charge in the message, and if the FCU/battery monitor provides only capacity and percent, compute charge.

None of these are perfect. (1) has redundant data which can cause problems. It's easy for someone to write a node which expects capacity and percent, for example, and have it work for a while with their hardware, but then fail unexpectedly when they use a new device that fills in capacity and charge but not percent. With (2) and (3) there is a possibility that the message will have to compute a value based on others instead of exposing the value that is actually known.

My personal preference is (3) because from an engineering perspective charge ought to be the lower level value and percent is a computed value that downstream nodes can be expected to calculate. From a practical standpoint, however, if most devices are going to output percent and not charge, I could see (2) being a reasonable choice as well. Personally I quite dislike option (1) because of the potential for bugs.

jack-oquin:
1. seems reasonable to me, too.

tfoote:
I've been tossing this over in my head. 

1) makes downstream users code slightly simpler. I think we could make the percentage required, and the charges/capacity optional. 
2 + 3) require the downstream consumer to compute a potentially measured value. 

I agree with a preference for 3 over 2. But I'm torn that I think 1) with percentage required will be easier for the consumer. And as there will likely be many more consumers than battery drivers I'd be in favor of leaning that way, and making sure that we just make sure to double check the battery drivers logic.

stonier:
+1 for putting computational logic into as few places as possible, i.e. the drivers.

Especially as it seems this message is targeted as a generic interface to the broadest group of battery/smart battery consumers as possible. If people require light messages at high frequencies, this should be a smaller use case which can be supported in a custom/alternative manner.

tfoote:
I agree with @stonier that focusing on the users code should be a priority. And as @jonbinney pointed out there is a risk of bugs, but I expect the number of battery drivers to be limited hopefully keeping the potential impact low. I agree too that the bandwidth consumption should not be a problem for most use cases. 

As such I think that the current proposal is in line with 1) from above. I'd like to propose going ahead with the current form. Is there anything else people would like to talk about before considering merging this?

jack-oquin:
My feeling is that the typical use case should publish at most a few times a second, so bandwidth would be irrelevant.

There may be other use cases where it _is_ relevant, but I can't think of any right now.

stonier:
I had one other topic I brought up elsewhere with Tully concerning the addition of a `charging source` variable to go along with the `charging state` that is present here.

We use docking stations and jacks (i.e. just plug in the cable) for all of our robots and so we often have higher level logic making decisions based on the charging source. I do not think this is the majority use case though and it is not something we need to expose all the way through to user interfaces (just control programs), so we're currently leaning towards using this for battery state and a separate `ChargingState`.

Does anyone else use/require charging source along with the charging state?

tfoote:
I think keeping the charging state separate makes sense as the battery itself doesn't really know that. That information comes from the charger/power controller board which is muxing it into the smart batteries inputs, whereas the battery just sees the input voltage higher than the output voltage.

justinthomas:
Only thing I might suggest is including the chemistry type so that if individual cell voltages are provided, the user can determine if one cell in particular is too high or low.

meyerj:
What exactly is the meaning of the charge_state field? Is this
1. the _intended_ charge state (e.g. commanded by the user or reported by some charger logic) or
2. the _real_ charge state, an abstraction of the current (and charge) information?

In the first case I would suggest to add something like `NONE` or `IDLE` as the default (value 0) and use `DISCHARGING` only if the battery is connected to an internal or external discharger (which might be required from time to time to have a good charge estimate aka recalibration). In the second case, if the charge_state simply toggles based on the current being negative or positive it would be somehow redundant to the current and charge fields, given that those fields are set. When exactly would a battery driver set it to `CHARGED` and when back to `DISCHARGING`?

In both cases I would suggest to stay close to the ACPI charging states in the Linux kernel (see [drivers/acpi/battery.c](http://lxr.free-electrons.com/source/drivers/acpi/battery.c?v=4.2#L60) or directly in the [ACPI specs](http://www.acpi.info/DOWNLOADS/ACPIspec50.pdf) [1]), as a driver that reads the battery state of a laptop from `/proc/acpi/battery/BATx/state` would probably be a kind of standard driver that publishes this message:

```
unit8 IDLE = 0            # or UNREPORTED / UNKNOWN?
uint8 DISCHARGING = 1
uint8 CHARGING = 2
uint8 CRITICAL = 4
uint8 CHARGED = 8
uint8 UNREPORTED = 128    # is this needed???
```

Even though `CHARGED` is not a defined as a constant in battery.c, it can be a value of the `charging state:` field in the textual output as reported in the proc file system (see [battery.c:750](http://lxr.free-electrons.com/source/drivers/acpi/battery.c?v=4.2#L850)). Obviously they are systems where ACPI might set both, the discharging and charging flags, and the `CRITICAL` flag is independent, so `charge_state` should be a bit field. I am not sure whether there is a need for a separate `UNREPORTED` value or if the value 0 represents an unknown/unreported charge state, but there might be a difference between a battery being half-full and not charging or discharging (e.g. while the robot is AC powered) and a completely unknown battery state.

Also for other fields the [struct acpi_battery](http://lxr.free-electrons.com/source/drivers/acpi/battery.c?v=4.2#L118) in battery.c and the [linux/power_supply.h](http://lxr.free-electrons.com/source/include/linux/power_supply.h?v=4.2) header might be worth a look, e.g. for the battery technology/chemistry.

[1] http://www.acpi.info/DOWNLOADS/ACPIspec50.pdf p. 505

tfoote:
> What exactly is the meaning of the charge_state field? Is this
> 1. the intended charge state (e.g. commanded by the user or reported by some charger logic) or
> 2. the real charge state, an abstraction of the current (and charge) information?

I would state that it should be reporting the real state. 

> In the first case I would suggest to add something like NONE or IDLE as the default (value 0) and use DISCHARGING only if the battery is connected to an internal or external discharger (which might be required from time to time to have a good charge estimate aka recalibration). In the second case, if the charge_state simply toggles based on the current being negative or positive it would be somehow redundant to the current and charge fields, given that those fields are set. When exactly would a battery driver set it to CHARGED and when back to DISCHARGING?

I would agree that the real state should be basically introspectable from the voltage and current measurements. However, the current is currently defined as optional which means that this is not entirely redundant. Also in smart batteries this charge state should reflect the state of the internal charge controller which is otherwise not visible to the system. Depending on the situation it might make sense to detect that the internal charging logic and the current/voltage combinations are in conflict, which might detect a malfunction or something out of range. 

> In both cases I would suggest to stay close to the ACPI charging states in the Linux kernel (see drivers/acpi/battery.c or directly in the ACPI specs [1]), as a driver that reads the battery state of a laptop from /proc/acpi/battery/BATx/state would probably be a kind of standard driver that publishes this message:
> 
> unit8 IDLE = 0            # or UNREPORTED / UNKNOWN?
> uint8 DISCHARGING = 1
> uint8 CHARGING = 2
> uint8 CRITICAL = 4
> uint8 CHARGED = 8
> uint8 UNREPORTED = 128    # is this needed???
> Even though CHARGED is not a defined as a constant in battery.c, it can be a value of the charging state: field in the textual output as reported in the proc file system (see battery.c:750). Obviously they are systems where ACPI might set both, the discharging and charging flags, and the CRITICAL flag is independent, so charge_state should be a bit field. I am not sure whether there is a need for a separate UNREPORTED value or if the value 0 represents an unknown/unreported charge state, but there might be a difference between a battery being half-full and not charging or discharging (e.g. while the robot is AC powered) and a completely unknown battery state.
> 
> Also for other fields the struct acpi_battery in battery.c and the linux/power_supply.h header might be worth a look, e.g. for the battery technology/chemistry.
> [1] http://www.acpi.info/DOWNLOADS/ACPIspec50.pdf p. 505

Thanks for the pointers to the acpi structs. I think it would make sense to reuse their enums for charging state and chemistry[L35](http://lxr.free-electrons.com/source/include/linux/power_supply.h#L35) [L62](http://lxr.free-electrons.com/source/include/linux/power_supply.h#L62). The other one from that file which I think we might want to consider adding is the health metrics.[L50](http://lxr.free-electrons.com/source/include/linux/power_supply.h#L50) They have a good list of failure modes which we could add as an enum too.  Each one of these could be a uint8.

tfoote:
From the input's I've updated the struct to support the status, health, and technology as used in the power_supply acpi elements. Please take a final look at this as I think we've reached a consensus.

sensor_msgs/msg/BatteryState.msg
```diff
@@ -0,0 +1,43 @@
+# Power supply status parallel to defines in power_supply.h
+uint8 POWER_SUPPLY_STATUS_UNKNOWN = 0
+uint8 POWER_SUPPLY_STATUS_CHARGING = 1
+uint8 POWER_SUPPLY_STATUS_DISCHARGING = 2
+uint8 POWER_SUPPLY_STATUS_NOT_CHARGING = 3
+uint8 POWER_SUPPLY_STATUS_FULL = 4
+
+# Power supply health parallel to defines in power_supply.h
+uint8 POWER_SUPPLY_health_UNKNOWN = 0
```
jonbinney: Why is "health" lowercase here?

tfoote: Search and replace failure. Fixed.

sensor_msgs/msg/BatteryState.msg
```diff
@@ -0,0 +1,43 @@
+# Power supply status parallel to defines in power_supply.h
```
jonbinney: Should probably specify that the `power_supply.h` header is from the linux sourcecode, since some people may not assume that. Might also be worth specifying the version of the linux sourcecode it has been borrowed from.

sensor_msgs/msg/BatteryState.msg
```diff
@@ -0,0 +1,43 @@
...
+uint8 POWER_SUPPLY_HEALTH_DEAD = 3
+uint8 POWER_SUPPLY_HEALTH_OVERVOLTAGE = 4
+uint8 POWER_SUPPLY_HEALTH_UNSPEC_FAILURE = 5
+uint8 POWER_SUPPLY_HEALTH_COLD = 6
+uint8 POWER_SUPPLY_HEALTH_WATCHDOG_TIMER_EXPIRE = 7
+uint8 POWER_SUPPLY_HEALTH_SAFETY_TIMER_EXPIRE = 8
+
+# Power supply technology (chemistry) parallel to defines in power_supply.h
+uint8 POWER_SUPPLY_TECHNOLOGY_UNKNOWN = 0
+uint8 POWER_SUPPLY_TECHNOLOGY_NIMH = 1
```
jonbinney, line 27: Capitalization is inconsistent here. "NIMH" is all caps, despite being "NiMH" in power_supply.h, whereas "LiFe" has lowercase letters here. Should probably either always follow ROS convention (all caps for constants) or always follow power_supply.h capitalization. Personally I tend to lean toward keeping the ROS convention.

jonbinney:
Other than a few minor style issues I've commented on, LGTM

sensor_msgs/msg/BatteryState.msg
tfoote: Comment added

sensor_msgs/msg/BatteryState.msg
tfoote, line 27: I was going back and forth to copy directly or capitalize. I'll capitalize them all to keep our codebase cleaner.

tfoote:
Thanks for the feedback @jonbinney I've resolved your comments. 

Does anyone else have any outstanding concerns?

stonier:
Looks good to me :neckbeard:

bit-pirate:
:+1:

jack-oquin:
+1

tfoote:
Thanks everyone for all the feedback. It looks like we've reached a solution. I'm going to merge it now and will cherry pick into jade.

PR #79, opened 2015-09-08
diagnostic_msgs: Add services for dynamic addition, removal and listing of analyzers

heuristicus:
See ros/diagnostics/pull/41 for the code which uses these new srvs.

PR #80, opened 2015-10-02
diagnostic_msgs: Add messages for service used to add diagnostics to aggregator

heuristicus:
See ros/diagnostics#43 for the corresponding changes to the aggregator.

### DIFF
--- a/diagnostic_msgs/srv/AddDiagnostics.srv
+++ b/diagnostic_msgs/srv/AddDiagnostics.srv
@@ -1,20 +1,22 @@
 # This service is used as a request to create a bond between the diagnostic
 # aggregator and the node sending the request.
 
-# The namespace into which parameters for the initialization of analyzers in the
-# diagnostic aggregator have been loaded. This also defines the name of the bond
-# which will be created by the aggregator. A bond with that name should already
-# exist on the requesting side before this request is sent. The value should be
-# a global name (i.e. /my/name/space), not a relative (my/name/space) or private
-# (~my/name/space) name. Analyzers will not be added if a non-global name is
-# used. The call will also fail if the namespace contains parameters that follow
-# a namespace structure that does not conform to that expected by the analyzer
-# definitions. See
+# The load_namespace parameter defines the namespace where parameters for the
+# initialization of analyzers in the diagnostic aggregator have been loaded.
+# This also defines the name of the bond which will be created by the
+# aggregator. A bond with that name should already exist on the requesting side
+# before this request is sent. The value should be a global name (i.e.
+# /my/name/space), not a relative (my/name/space) or private (~my/name/space)
+# name. Analyzers will not be added if a non-global name is used. The call will
+# also fail if the namespace contains parameters that follow a namespace
+# structure that does not conform to that expected by the analyzer definitions.
+# See
 # http://wiki.ros.org/diagnostics/Tutorials/Configuring%20Diagnostic%20Aggregators
 # and http://wiki.ros.org/diagnostics/Tutorials/Using%20the%20GenericAnalyzer
-# for examples of the expected yaml structure.
+# for examples of the structure of yaml files which are expected to have been
+# loaded into the namespace.
 string load_namespace
----
+ ---
 
 # True if diagnostic aggregator was updated with new diagnostics, False
 # otherwise. A false return value means that either there is a bond in the
--- a/diagnostic_msgs/srv/AddDiagnostics.srv
+++ b/diagnostic_msgs/srv/AddDiagnostics.srv
@@ -1,0 +1,6 @@
+# Namespace containing parameters to use to set up additional diagnostics, in
+# the format expected by the diagnostic aggregator.
+string load_namespace 
+---
+bool success # True if diagnostic aggregator was updated with new diagnostics
+string message # Message with information about success or failure
--- a/diagnostic_msgs/srv/AddDiagnostics.srv
+++ b/diagnostic_msgs/srv/AddDiagnostics.srv
@@ -1,6 +1,26 @@
-# Namespace containing parameters to use to set up additional diagnostics, in
-# the format expected by the diagnostic aggregator.
-string load_namespace 
+# This service is used as a request to create a bond between the diagnostic
+# aggregator and the node sending the request.
+
+# The namespace into which parameters for the initialization of analyzers in the
+# diagnostic aggregator have been loaded. This also defines the name of the bond
+# which will be created by the aggregator. A bond with that name should already
+# exist on the requesting side before this request is sent. The value should be
+# a global name (i.e. /my/name/space), not a relative (my/name/space) or private
+# (~my/name/space) name. Analyzers will not be added if a non-global name is
+# used. The call will also fail if the namespace contains parameters that follow
+# a namespace structure that does not conform to that expected by the analyzer
+# definitions. See
+# http://wiki.ros.org/diagnostics/Tutorials/Configuring%20Diagnostic%20Aggregators
+# and http://wiki.ros.org/diagnostics/Tutorials/Using%20the%20GenericAnalyzer
+# for examples of the expected yaml structure.
+string load_namespace
 ---
-bool success # True if diagnostic aggregator was updated with new diagnostics
-string message # Message with information about success or failure
+
+# True if diagnostic aggregator was updated with new diagnostics, False
+# otherwise. A false return value means that either there is a bond in the
+# aggregator which already used the requested namespace, or the initialization
+# of analyzers failed.
+bool success
+
+# Message with additional information about the success or failure
+string message
--- a/diagnostic_msgs/srv/AddDiagnostics.srv
+++ b/diagnostic_msgs/srv/AddDiagnostics.srv
@@ -1,22 +1,21 @@
-# This service is used as a request to create a bond between the diagnostic
-# aggregator and the node sending the request.
+# This service is used as part of the process for loading analyzers at runtime,
+# and should be used by a loader script or program, not as a standalone service.
+# Information about dynamic addition of analyzers can be found at
+# http://wiki.ros.org/diagnostics/Tutorials/Adding%20Analyzers%20at%20Runtime
 
 # The load_namespace parameter defines the namespace where parameters for the
-# initialization of analyzers in the diagnostic aggregator have been loaded.
-# This also defines the name of the bond which will be created by the
-# aggregator. A bond with that name should already exist on the requesting side
-# before this request is sent. The value should be a global name (i.e.
-# /my/name/space), not a relative (my/name/space) or private (~my/name/space)
-# name. Analyzers will not be added if a non-global name is used. The call will
-# also fail if the namespace contains parameters that follow a namespace
-# structure that does not conform to that expected by the analyzer definitions.
-# See
+# initialization of analyzers in the diagnostic aggregator have been loaded. The
+# value should be a global name (i.e. /my/name/space), not a relative
+# (my/name/space) or private (~my/name/space) name. Analyzers will not be added
+# if a non-global name is used. The call will also fail if the namespace
+# contains parameters that follow a namespace structure that does not conform to
+# that expected by the analyzer definitions. See
 # http://wiki.ros.org/diagnostics/Tutorials/Configuring%20Diagnostic%20Aggregators
 # and http://wiki.ros.org/diagnostics/Tutorials/Using%20the%20GenericAnalyzer
 # for examples of the structure of yaml files which are expected to have been
 # loaded into the namespace.
 string load_namespace
- ---
+---
 
 # True if diagnostic aggregator was updated with new diagnostics, False
 # otherwise. A false return value means that either there is a bond in the
--- a/diagnostic_msgs/CMakeLists.txt
+++ b/diagnostic_msgs/CMakeLists.txt
@@ -13,6 +13,7 @@ add_message_files(
 add_service_files(
   DIRECTORY srv
   FILES
+  AddDiagnostics.srv
   SelfTest.srv)
 
 generate_messages(DEPENDENCIES std_msgs)


tfoote:
This has zero documentation and looks like a package local feature. Does it need to be in common_msgs. It looks like it's expected to be called and interact with one script and the diagonstic analyzer thus separating the message out does not make sense.

heuristicus:
I've added some comments to the `.srv`.

The current implementation is a package local feature, but that doesn't preclude its use outside of the package, for example in a node. Such an implementation could be useful for users who need more control over when diagnostics come up and down.

stonier:
Right now, yes, it's package local. However, if you keep it locked up inside the package sources, it precludes the ability of someone wanting to write a java, or windows/cross compiled c++ equivalent to that script (since they would then need to compile the core diagnostics package which is likely be impossible and an unnecessary dependency regardless). I've had this headache so often in the past.

Anything that is merely an interaction with another node should (if there is no attached cost) be freely usable without having to compile the sources. Otherwise we automatically bound what future uses people may have for them. And when this boundary is realised, there is a large `api change` barrier to overcome when the message has to move across packages.

stonier:
This is a `work in progress`. I'll get @heuristicus to update here if/when this is ready to roll.

heuristicus:
The changes to the diagnostic aggregator are ready to merge, and require these messages for functionality. I still think that these shouldn't be package local, to allow custom interactions with the aggregator from outside the package.

tfoote:
if it's been developed and seems to have consensus that the functionality is reasonable for the diagnostic aggregator we can go ahead and add it here. However to add it to common_msgs it will need to be reviewed to the level of everything else here.

Right now it's a message basically containing an undocumented datstructure which is serialized into a string. It needs to take advantage of the datastructure available in messages. And it needs to be self documenting such that someone else could pick up and parse the message without access to the aggregator source. 

Also does this need the other services in  #79?

trainman419:
This is independent of #79; the API has be rewritten to use add+bond instead of separate services for listing, adding and removing aggregators.

heuristicus:
I'm not sure what you mean by something undocumented being serialized. The `AddDiagnostics` service is used as a request to create a bond between the diagnostic aggregator and the node sending the request. The `string` parameter in the request is just the namespace from which the aggregator should read parameters for setting up its analyzers. As an example, the script that is meant to be the main way of adding analyzers sends across the string `rospy.resolve_name(rospy.get_name())`. There's no serialization going on anywhere.

tfoote:
I was just reading the documentation "in the format expected by the diagnostic aggregator." From reading that comment I have no clue what data is inside the string and it implies a non-standard format in citing an external source. I shouldn't have to read into other sources. 

Please update the comments for the message to include everything you are explaining in this thread. A future user will need to be able to understand the message and it's use by reading the message definition with comments. 

Make sure to also cover what are valid values too. Such as does it need to be absolute or can it be relative? If it's relative what is it relative to? That sort of thing. Please read through some of the [sensor_msgs](https://github.com/ros/common_msgs/tree/jade-devel/sensor_msgs/msg) for examples of the documentation I'd like to see. 

On a higher level I have a few comments on the design. First it seems like this message is sending a reference to other data which the remote end will then load. Why is the actual data not sent in the request? What are the semantics for reloading? Are there race conditions for loading the expected parameters? Can you query for what the state of the loaded aggregators?

stonier:
**As it stands**

I think the comment should be at least more explicit about being a namespace for parameters on the `ros parameter server` - that's not exactly clear. Whether absolute or relative is also a good point of view. It would also be good to have a link to somewhere explaining what data structure should be found on the parameter server.

**Is it the right way?**

The typical case for Michal's example is that there is a yaml configuration which is loaded with the node that sends this request to the diagnostics aggregator. The natural way to do this in ros is via the ros parameter server and roslaunch makes this a one-liner in xml.

So the problem is how to get this information to the dynamic aggregator. Populating the message explicitly seems like a redundant step when the information is publicly available to every node, but certainly possible.

Are there desirable reasons/conventions for not using the ros parameter server in this way? 

It does assume uniqueness of that namespace and that this namespace is not used by anyone else. It also loads into the loading node's private namespace, and you could say that it is an abuse of the notion of a private namespace, particularly when the notion of privacy is not really enforced.

Even if we did wish to populate the message type explicitly, it is probably not possible to create a message type that covers all of the variables thrown into an aggregator yaml. Yaml/ros param is the current way things are done, and msg's are too limited to handle this. e.g. it would have to create a message type covering these variables (and there are many more)

```
analyzers:
  navi:
    type: diagnostic_aggregator/AnalyzerGroup
    path: Navigation
    analyzers:
      dslam:
        type: diagnostic_aggregator/GenericAnalyzer
        path: DSlam
        timeout: 5.0
        num_items: 11
        find_and_remove_prefix: 'gopher dslam'
      marco_polo:
        type: diagnostic_aggregator/GenericAnalyzer
        path: MarcoPolo
        timeout: -1 # never stale
        num_items: 1
        find_and_remove_prefix: 'marco_polo::'
```

So if we can't create a ros message explicitly and don't want to use the ros parameter server, the only options left are to either send a yaml string or filename to the diagnostics aggregator.

I'm not sure either of those options are all that palatable either. The filename option has similar concerns as those for the parameter server.

heuristicus:
Part of the reason that I used the namespace passing is because originally I didn't want there to be any communication between the aggregator and any node adding diagnostics, other than via the bond. This turned out to not be possible because of the need to match names, and the difficulty/messiness of getting the aggregator to find incoming bonds. I already feel like I'm abusing the bond name a little, and while passing a yaml through that is possible, it's definitely not how it was intended to be used. 

I think that actually it could be possible to do away with the service and use only bonds, but that would require the aggregator to periodically check incoming messages on the bond topic and create bonds which it hasn't got in its list yet. I prefer the current way, though, since there is more clearly a request for the addition.

Passing a namespace rather than all of the data has the benefit that the user can load any number of analyzer definitions specified in different files into that namespace, and they can all be read at the same time. I suppose that's not really an argument against just defining which files to load, reading them, and then passing all the data, but doing that just feels messy. The parameters have to be loaded into a namespace at some point anyway, so why not just do it where it's most convenient?

I'm not quite sure what you mean by semantics for reloading. Any parameter data defining analyzers that is loaded into the node's private namespace is not deleted when that node shuts down. If the same node is brought up again with a different analyzer specifications, then that might cause some problems. We haven't yet discussed what should be done in this case.

It is likely that there is some kind of race condition for parameter loading, but I think it would be hard to get it to manifest itself. The parameters are read only when a request to add the analyzers is sent. If the parameters haven't been loaded yet, then the analyzer initialisation fails and the request returns failure. If you are loading the parameters in another thread at the same time as making the request, it might be the case that they won't be fully loaded by the time the request goes through. The current script is uploading parameters to the server before the request is made. I don't know the internals of how that works, but I'm assuming that once the call has been made those parameters are available on the server. Perhaps this is an incorrect assumption, but I haven't encountered any issues with loading parameters.

There is no way of querying the state of the loaded analyzers. Initially this was going to be included, when the plan was to allow anyone to load/unload analyzers, but once we switched to using bonds it didn't seem necessary.

@trainman419 do you have any comments on this?

tfoote:
> So the problem is how to get this information to the dynamic aggregator. Populating the message explicitly seems like a redundant step when the information is publicly available to every node, but certainly possible.
> 
> Are there desirable reasons/conventions for not using the ros parameter server in this way?

The reason I'm asking is that the parameter server is not a good place to store state. It's not logged, it's not atomic and there's no real 'ownership' of the data. In the recommended default case where the launch file simply loads the parameters up, then you pass the namespace to the launcher and it reads them and loads them that's fine. However, if someone who's not familiar with the system starts poking around and finds all these parameters, they may try to tweak them. But when they tweak them there will be no effect since the node has loaded them in the past. Another use case which is very hard to debug in this design is if you forgot to request loading a namespace, I don't see any way to determine that the namespace was not loaded. You'll find the parameters there, but the analyzer won't be using them and there's no way to tell why. 

I don't see anything in that structure that cannot be relatively easily contained in a message. We don't support dicts, but a good proxy for most cases is a list of named items. 

Subanalyzers.msg

```
string name
string type
string path
float64 timeout
uint16 num_items
string find_and_remove_prefix
```

Analyzer.msg

```
string name
string type
string path
Subanalyzers analyzers[]
```

You could then create getter and setter actions or services using these primatives.

trainman419:
Part of the trouble here is that the analyzer structure is not nearly as concrete as your message types suggest. In particular:
- Analyzers can be arbitrarily grouped and nested
- New analyzers can specify new parameters

Together, this basically means that the only appropriate data structure for representing the configuration data for an analyzer is an arbitrarily nested dictionary, and representing that in a ROS message type is difficult and cumbersome.

heuristicus:
> someone who's not familiar with the system starts poking around and finds all these parameters, they may try to tweak them

This applies to the diagnostic aggregator itself as well. The parameters are loaded to the parameter server in the launch file and then read when the aggregator initialises. If someone is using the aggregator at all then they should already know that modifying the parameters there has no effect. The dynamic addition behaves in the same way.

> Another use case which is very hard to debug in this design is if you forgot to request loading a namespace

Forgetting to make a request will result in the unpaired bond timing out due to the sister not coming up in time, so if there is adequate handling on the end which is making the load request, this should be pretty obvious.

We've come up with a couple of possible ways of continuing:

##### Send yaml strings to the aggregator

Instead of passing a reference to the namespace where the parameters are already loaded, we send the full yaml of the analyzer definitions to the aggregator, which then parses that and loads it into its own private namespace. This has the benefit of not abusing the idea of a private namespace, and also means that the message contains all the required information.

##### Send a filename to the aggregator

Instead of reading the file before it gets to the aggregator, we just send the path to a file containing analyzer definitions, and the aggregator loads it up.

##### Stop using a service, use only bonds instead

We stop using this service and perform the linking purely with bonds. The reason for the service is to allow the aggregator to know that there is someone wanting to add analyzers, and also to pass information about the parameters for the new analyzers. The service is probably not strictly necessary, but makes things more explicit. Instead, we could have a subscriber in the aggregator which listens on the bond topic for nodes with names that it hasn't seen before, and creates sister bonds to link to them. 

This bond-only approach would use the bond names to define what the aggregator is loading from; a filename, namespace or yaml string. This is an abuse of the bond name, which isn't really intended for data passing, though. It does mean that the service is unnecessary and reduces the number of steps that need to be taken to add analyzers.

stonier:
I'd like to bump this somehow - this single non-critical addition has been struggling to forge a way ahead for a considerably long period of time. So much so that it is starting to feel like the effort to contribute (for this item) is exceeding the effort that can be afforded.

@heuristicus has made some suggestions on the alternatives that are available. All of them, including the current implementation, do the job, and all have some less desirable cons. However, none are showstoppers and the benefits outweigh anything that has been mentioned above. Rather than @hueristicus spending time guessing and implementing again, @tfoote could you nail down an option that is acceptable and he can push this forward to closure?

My own two cents:

I guess the most important thing from this discussion is a design pattern for how external processes  can pass/withdraw configuration to a running node.

Right now, we are using bonded nodes for the up/down signalling they confer and the configuration has naturally moved to the ros param server since neither dynamic reconfigure nor ros messaging can  provide. It also allows reuse of code inside the aggregator for the purposes of configuration. Yes, the ros param server has limitations, but it is the default way we configure most ros nodes. Atomicity is not a problem, roslaunch loads them up before it executes nodes. State also - most ros nodes follow the pattern of loading up parameters at launch time and ignoring the parameters thereafter (including the diagnostics aggregator). This pattern is already too established and the param server does not have richer mechanisms to handle this problem anyway.

Let's get this going before we all get too old :neckbeard:

tfoote:
I agree that this is not something for which there's a clearly great option. The current implementation is something that does the job but has a lot of drawbacks. However any of the alternatives have a similar level of drawbacks because we don't have more structured interactions with the parameter server and the data is not structured such that it can go into messages. 

Pending a large rework of the data I think we can take this service into common_msgs if we just really document it with all the implications etc that are discussed in this thread. A user should be able to read the message documentation and understand how to craft their own copy or interpret it by reading the message documentation. Right now it's just a one liner which does not even link to the expected format of the target.

heuristicus:
I've added some information to the comments.

diagnostic_msgs/srv/AddDiagnostics.srv
```diff
@@ -0,0 +1,26 @@
+# This service is used as a request to create a bond between the diagnostic
+# aggregator and the node sending the request.
+
+# The namespace into which parameters for the initialization of analyzers in the
```
trainman419: This is quite a block of text. Maybe reword this as:

> The `load_namespace` parameter defines the namespace where parameters for the initialization of analyzers in the

diagnostic_msgs/srv/AddDiagnostics.srv
```diff
@@ -0,0 +1,26 @@
...
+# which will be created by the aggregator. A bond with that name should already
+# exist on the requesting side before this request is sent. The value should be
+# a global name (i.e. /my/name/space), not a relative (my/name/space) or private
+# (~my/name/space) name. Analyzers will not be added if a non-global name is
+# used. The call will also fail if the namespace contains parameters that follow
+# a namespace structure that does not conform to that expected by the analyzer
+# definitions. See
+# http://wiki.ros.org/diagnostics/Tutorials/Configuring%20Diagnostic%20Aggregators
+# and http://wiki.ros.org/diagnostics/Tutorials/Using%20the%20GenericAnalyzer
+# for examples of the expected yaml structure.
```
trainman419: s/yaml/parameter/ (data on the parameter server is not yaml)

diagnostic_msgs/srv/AddDiagnostics.srv
```diff
@@ -0,0 +1,28 @@
+# This service is used as a request to create a bond between the diagnostic
+# aggregator and the node sending the request.
+
```
stonier, line 5: This could be misleading. As it is, it seems like you could load some params and then call this service on the command line to do the job.

The reality however, is you need a loader script/program that is responsible for bonding with _and_ transferring information via this service to the aggregator, such as the python loader you have contributed. Could you reword this to that effect, perhaps with a link to the tutorial for adding diagnostics at runtime?

trainman419:
@tully how do the proposed documentation changes look? This has been open for quite a while and it would be nice to get it merged.

tfoote:
@heuristicus Thanks for adding the extra documentation. I'll merge it in and work to do a coordinated release with @dirk-thomas ad I think there's an upcoming catkin release requiring a rebuild.

trainman419:
I'm going to update diagnostics so that it explicitly relies on the next version of common_msgs; will this be released in version 1.11.9 ?

trainman419:
Poke. Is this included in common_msgs 1.11.9?

I'd like to get a release of the updated version of diagnostics rolling asap, so that we can find any blocking bugs early in the build process, and I'd like to have a versioned dependency on common_msgs. How should I formulate that dependency?

tfoote:
Yes it's in [1.11.9](https://github.com/ros/rosdistro/pull/10558) and [1.12.4](https://github.com/ros/rosdistro/pull/10559) once the PRs are merged you can add a >= dependency on those versions.   Note you could queue up the release for simultaneous release too. Just make a comment about the dependency on the appropriate other release.

PR #82, opened 2015-12-21
remove boost dependency

### LINKED ISSUE #81: sensor_msgs dependency on boost
The sensor_msgs package depends on boost because `boost::str()` and `boost::format()` are called to format an error message in [point_cloud_conversion.h:120](https://github.com/ros/common_msgs/blob/89788101b21aec550ef4d0b93c7f668e9e8d19c0/sensor_msgs/include/sensor_msgs/point_cloud_conversion.h#L120), but...
- boost is not explicitly declared as a dependency (of type `<run_depend>` or `<build_export_depend>`?) in package.xml, only indirectly through message_runtime.
- point_cloud_conversion.h is lacking an include directive for `<boost/format.hpp>` and includes this header only indirectly via `ros/time.h`:
  
  ```
  . /opt/ros/indigo/include/sensor_msgs/point_cloud_conversion.h
  .. /opt/ros/indigo/include/sensor_msgs/PointCloud.h
  ... /opt/ros/indigo/include/ros/serialization.h
  .... /opt/ros/indigo/include/ros/time.h
  ..... /usr/include/boost/math/special_functions/round.hpp
  ...... /usr/include/boost/math/policies/error_handling.hpp
  ....... /usr/include/boost/format.hpp
  ```
- `convertPointCloud2ToPointCloud()` should return false after printing the message to stderr in line 120.

This is not directly related to those three issues, but we spent some time finding a problem while building sensor_msgs and dependents on a RedHat system with cmake 2.8.9, because sensor_msgs picked up and exported wrong Boost library paths in its generated cmake config if it was compiled without the `-DBoost_NO_BOOST_CMAKE=ON` argument (see http://stackoverflow.com/questions/9948375/cmake-find-package-succeeds-but-returns-wrong-path).

tfoote:
fixes #81

### DIFF
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 2.8.3)
 project(sensor_msgs)
 
 find_package(catkin REQUIRED COMPONENTS geometry_msgs message_generation std_msgs)
-# We want boost/format.hpp, which isn't in its own component.
-find_package(Boost REQUIRED)
 
 # For point_cloud2.py
 catkin_python_setup()
@@ -49,8 +47,7 @@ generate_messages(DEPENDENCIES geometry_msgs std_msgs)
 
 catkin_package(
   INCLUDE_DIRS include
-  CATKIN_DEPENDS geometry_msgs message_runtime std_msgs
-  DEPENDS Boost)
+  CATKIN_DEPENDS geometry_msgs message_runtime std_msgs)
 
 install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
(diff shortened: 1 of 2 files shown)


tfoote:
@meyerj @vrabaud Does this look reasonable as a fix for #81 ?

meyerj:
`convertPointCloud2ToPointCloud()` should definitely return false after printing the error message, or the result is undefined.

Otherwise this patch lgtm, because it removes the Boost dependency without changing the actual output, but I am still not sure if writing to std::cerr is a good idea in such a low-level conversion function (that largely depends on how it is used in user code). At least the header should `#include <iostream>` then. Alternatives would be to add a dependency to rosconsole and use the `ROS_ERROR()` macro or to throw a `std::runtime_error` instead and leave it to the caller to catch the exception.

jack-oquin:
I agree about the `cerr` output being undesirable at this level. 

Raising an exception is probably better.

vrabaud:
the only solution we can afford now without breaking the API / behavior is that PR + proper return as mentioned by @meyerj 

Ideal behavior to me: exception and / or ros_console log + sensor_msgs_helper package (split the current package between msg+srv and code).

tfoote:
Thanks for the reviews. I've fixed the return code and squashed.

tfoote:
cherry-picked to jade f24f24c248dcb36eb2deba4ffbd3d2c9b86d69d1

PR #83, opened 2016-02-23
Add diagnostics service

stonier:
Squashed changes from https://github.com/ros/common_msgs/pull/80 into jade-devel.

@trainman419 this one will be necessary for building on jade.


trainman419:
Thanks for the heads-up @stonier !

tfoote:
Thanks for the forward port.

PR #92, opened 2016-09-30
Changes the MODIFY value to 1 in Marker.msg struct

carlosjoserg:
Just found it while reading the docs.-

I don't know when it'd be useful to have a MODIFY action type, but I guess it should have a different value at least, right?

Probably nobody ever used this functionality before... maybe it was already expected that ADD and MODIFY should behave the same?

PR-ing just in case


tfoote:
modify was deprecated and merged with add.
`action: 0 = add/modify, 1 = (deprecated), 2 = delete, New in Indigo 3 = deleteall` 

It's documented [here](http://wiki.ros.org/rviz/DisplayTypes/Marker).

carlosjoserg:
:+1: I missed that

tfoote:
No worries, it took me several searches to find that, but I had a guess what I was looking for.

PR #94, opened 2016-10-16
Add LoadMap service

### LINKED ISSUE #279: map_server: reload_map_from_file service?
Hi there,

I was wondering if it would be useful to implement a service on map_server to reload the map from a file. I think it would be beneficial in multi-floor environments where one needs to change the map due to the robot changing floors, instead of having to kill the  map_server node and starting it again with the map of the new floor.

Please let me know if this can be already achieved by some other means

sloretz:
The Load map service is to be handled by map_server. It will load (or reload) a map from the given file path when called.

This would allow completion of pull request ros-planning/navigation#461
Which would close ros-planning/navigation#279

### DIFF
--- a/nav_msgs/srv/LoadMap.srv
+++ b/nav_msgs/srv/LoadMap.srv
@@ -1,0 +1,3 @@
+string map_path
+---
+bool success


nav_msgs/srv/LoadMap.srv
```diff
@@ -0,0 +1,3 @@
+string map_path
```
tfoote, line 1: A path here is not portable. I would suggest that this be a set map method not a load map method. If you need to load a map from file, there can be a standalone program to load it and send it.

sloretz: It sounds like portability is an important quality of code in common_msgs. Given that map_server has two purposes: load a map from a file (it takes a file path as an argument when it is launched), and offer a map information on /map and /map_metadata topics,  are you saying you'd like to see those two purposes split?

I can refactor map_server to only offer the map, and create a separate loader node to give that map to map server. It would then be possible for other loaders to be created in support of other systems.

sloretz: Thinking about my original use case, if I gave map_server a SetMap service I still want a LoadMap service somewhere else.

The system has a large number of maps that can't feasibly be loaded at the same time. When an event calls for changing maps, code calls a LoadMap service to change the map that is being used for navigation and localization. 

Say there was a map_file_loader which loaded a map from a file and called SetMap on map_server. It could accept the path to a map via a param, but then code would change the map by launching the map_file_loader node, and that is a lot less convenient than a service call. The map_file_loader could stay running indefinitely and offer a LoadMap service, but that still means having a LoadMap.srv in nav_msgs. I could make the code to load a map from a file be part of the event handling code, but then others wouldn't get the benefit of it being open source.

Thoughts?

tfoote: Separating the loading functionality and the serving functionality would make sense to then allow alternative sources/updates. It will be good to have clearly defined interfaces, as It's likely that we will want to support new offline map formats to support https://github.com/googlecartographer

tfoote: A load map service definitely has use cases. However it should be abstracted on something like a map name as a resource that can be found on the system. Doing this can allow the loader to use a different backend, (file system, database, etc). And if you're able to load, you will also need to be able to at least list available resources. And the load map service will need more options for error codes than a boolean, to allow better error recovery. Knowing if the resources does not exist, versus a malformed datastructure, vs a runtime hiccup are cases that likely should be communicable.

jacobperron: I agree with @tfoote regarding separate SetMap and LoadMap services. I'm interested in this functionality and plan to create the definitions and implement both in [map_server](https://github.com/ros-planning/navigation). If this sounds reasonable you can expect a PR on Wednesday.

jacobperron: I just realized [SetMap](https://github.com/ros/common_msgs/blob/jade-devel/nav_msgs/srv/SetMap.srv) already exists, so that's one less service to define.

tfoote:
Replaced by #95

PR #95, opened 2016-11-02
Add LoadMap service

jacobperron:
The service is for loading a map as an [nav_msgs/OccupancyGrid](http://docs.ros.org/indigo/api/nav_msgs/html/msg/OccupancyGrid.html) given a resource ID. It also features a **result** member to encode errors. I tried to make the result codes generic enough for support of arbitrary backends and underlying data structures.   

This could be implemented in a *map_loader* node or directly in [map_server](https://github.com/ros-planning/navigation) seeing how it already "loads" a map on startup.

### DIFF
--- a/nav_msgs/CMakeLists.txt
+++ b/nav_msgs/CMakeLists.txt
@@ -17,7 +17,8 @@ add_service_files(
   FILES
   GetMap.srv
   GetPlan.srv
-  SetMap.srv)
+  SetMap.srv
+  LoadMap.srv)
 
 add_action_files(
   FILES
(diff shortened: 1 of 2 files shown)


jacobperron:
Attempt to resolve #94

tfoote:
Thanks for submitting this. It's updated the service to use a string id instead of a path. And having the result codes is a good addition. However it needs a very clear semantic of how to resolve the resource id such that alternative implementations would behave similarly.

jack-oquin:
Making it a string requires clearly defining the semantics. The obvious choice is a URL, with at least these options provided:
- `file:///path/to/local/file.yaml`
- `package://our_maps_package/maps/floor3.yaml`

Other nice-to-have schemes could include `http:`, `https:`, and remote `file:`.

doisyg:
Can we merge this? This is blocking for : https://github.com/ros-planning/navigation/pull/461

jacobperron:
Since I've lost access to the origin repository, I've opened a continuation PR https://github.com/ros/common_msgs/pull/152. I've made changes there to switch to URL semantics.

PR #96, opened 2016-11-04
actionlib_msgs: add_action_files takes absolute paths for DIRECTORY a…

mrjogo:
…rgument. This matches behavior of add_message_files and add_service_files. It implements in jade-devel the functionality of seemingly-abandoned PR ros/common_msgs#43.

Note that the .action file install path was updated to match messages and services, however that means that absolute DIRECTORY paths are installed relative to CATKIN_PACKAGE_SHARE_DESTINATION (ie, /usr/src/myactionfiles ends up in /opt/ros/jade/share/mypackage/usr/src/myactionfiles). Not sure if this is desired behavior or a bug.


tfoote (approved):
Thanks for following up the comments and creating the new pr.

PR #101, opened 2017-03-11
Proposed new messages to replace IMU message.

tfoote:
From https://groups.google.com/forum/#!topic/ros-sig-drivers/2NvgNBOjcFQ

Also connected to ros-infrastructure/rep#95

I'm opening this for more review and to restart the discussion. 

I see that we need a few things to be implemented before this can be merged. 

 * [ ] homogenize this with the REP 145 proposal, ros-infrastructure/rep#95
 * [ ] Implement some python methods to convert forward and backwards between the old IMU message and the trio of new messages. And a node that can used to convert each way.
 * [ ] Provide an IMU transport lib (in c++ and python) that will provide the necessary synchronization helper methods to handle the parallel transport in a way that keeps things synchronized.

### DIFF
--- a/sensor_msgs/msg/AngularVelocity.msg
+++ b/sensor_msgs/msg/AngularVelocity.msg
@@ -1,0 +1,4 @@
+std_msgs/Header header
+
+geometry_msgs/Vector3 angular_velocity
+float64[9] angular_velocity_covariance # Row major about x, y, z axes
--- a/sensor_msgs/msg/Attitude.msg
+++ b/sensor_msgs/msg/Attitude.msg
@@ -1,0 +1,14 @@
+# This message represents 3D attitude
+
+int8 ENU=1
+int8 NED=2
+int8 ENU_FLOATING=3
+int8 NED_FLOATING=4
+
+std_msgs/Header header
+
+geometry_msgs/Quaternion orientation
+float64[9] orientation_covariance # Row major about x, y, z axes
+
+uint8 reference_frame_type # NED, ENU, NED_FLOATING, ENU_FLOATING
+string sensor_frame # The coordinate frame in the system of the observation.
--- a/sensor_msgs/msg/LinearAcceleration.msg
+++ b/sensor_msgs/msg/LinearAcceleration.msg
@@ -1,0 +1,3 @@
+std_msgs/Header header
+geometry_msgs/Vector3 linear_acceleration
+float64[9] linear_acceleration_covariance # Row major x, y z


sensor_msgs/msg/Attitude.msg
```diff
@@ -0,0 +1,14 @@
+# This message represents 3D attitude
+
+int8 ENU=1
+int8 NED=2
+int8 ENU_FLOATING=3
+int8 NED_FLOATING=4
```
mikepurvis, line 6: There'd be a comment here explaining what these mean and what their relationship is to `sensor_frame`?

sensor_msgs/msg/Attitude.msg
```diff
@@ -0,0 +1,14 @@
...
+
+std_msgs/Header header
+
+geometry_msgs/Quaternion orientation
+float64[9] orientation_covariance # Row major about x, y, z axes
+
+uint8 reference_frame_type # NED, ENU, NED_FLOATING, ENU_FLOATING
+string sensor_frame # The coordinate frame in the system of the observation.
```
peci1, line 14: Isn't this duplicating info from header?

sensor_msgs/msg/LinearAcceleration.msg
```diff
@@ -0,0 +1,3 @@
+std_msgs/Header header
+geometry_msgs/Vector3 linear_acceleration
+float64[9] linear_acceleration_covariance # Row major x, y z
```
peci1, line 3: Maybe consider adding `bool is_gravity_compensated`? Defaulting to false sounds good as that is the old behavior.

sensor_msgs/msg/Attitude.msg
peci1, line 6: Define these right above `reference_frame_type` to make it more clear these belong to it?

peci1, line 6: Prefix the constants with `REFERENCE_FRAME_`?

sensor_msgs/msg/Attitude.msg
mintar: No. From https://groups.google.com/g/ros-sig-drivers/c/2NvgNBOjcFQ/m/LfG6fhFbIBwJ:

```text
Overview of frames for AHRS

/map # semi-global reference
|
---> ahrs_fix_ned/enu
|
---> odom
       | ---> base_link
                 |------> imu_link (the frame in which the ahrs sensor is mounted, possibly anywhere)
```

The `sensor_frame` should always be the frame of the IMU (`/imu_link` in this example). The way I understood the original proposal, the `header.stamp` should be the reference frame, like this:

* if `reference_frame_type == ENU` then `header.stamp = ahrs_fix_enu`
* if `reference_frame_type == NED` then `header.stamp = ahrs_fix_ned`
* if `reference_frame_type == ENU_FLOATING` then `header.stamp = ahrs_fix_enu_floating`
* if `reference_frame_type == NED_FLOATING` then `header.stamp = ahrs_fix_ned_floating`

... where the TF frames `ahrs_fix_enu_floating` and `ahrs_fix_ned_floating` don't exist (= are not connected to the `/map` frame). This is a bit weird, but the "floating" reference frames mean that the attitude is relative to some arbitrary reference frame (usually the startup pose of the sensor), and the transformation between `/map` and the startup pose is not known.

sensor_msgs/msg/AngularVelocity.msg
```diff
@@ -0,0 +1,4 @@
+std_msgs/Header header
+
+geometry_msgs/Vector3 angular_velocity
+float64[9] angular_velocity_covariance # Row major about x, y, z axes
```
mintar, line 4: We should probably copy the comment from the `sensor_msgs/Imu` message here:

https://github.com/ros/common_msgs/blob/20a833b56f9d7fd39655b8491a2ec1226d2639b3/sensor_msgs/msg/Imu.msg#L5-L8

sensor_msgs/msg/Attitude.msg
mintar: Something along these lines should probably added as a comment to this message.

sensor_msgs/msg/Attitude.msg
```diff
@@ -0,0 +1,14 @@
+# This message represents 3D attitude
+
+int8 ENU=1
+int8 NED=2
```
Timple, line 4: Why not enforce one of these?

Now all subscribers should handle this. I would like it better when there is only one definition within the ROS ecosystem.

peci1: Can you clarify how the enforcing should be done?

Timple: Stating in the message header-comment that ENU convention should be used.

ros-discourse:

https://discourse.ros.org/t/if-you-work-with-imu-data-share-sample-and-get-thanked/30203/4

sloretz:
Thank you for the PR!

PR #102, opened 2017-04-27
add mingration rule, copied from common_msgs-1.6

k-okada:
migration rule exists until common_msgs-1.6 (https://github.com/ros/common_msgs/blob/common_msgs-1.6/sensor_msgs/manifest.xml#L15) has been removed , when converting to catkin
may be this is rarely used, but I found we need this on http://wiki.ros.org/stereo_image_proc/Tutorials/ChoosingGoodStereoParameters


jack-oquin (approved):
Seems OK to me. Does catkin_lint approve?

k-okada:
> Does catkin_lint approve?

may be it' ok

```
k-okada@p40-yoga:~/catkin_ws/ws_migration/src/common_msgs/sensor_msgs$ catkin_lint -W2
sensor_msgs: test/CMakeLists.txt(2): error: missing test_depend on 'rosunit'
sensor_msgs: test/CMakeLists.txt(3): error: missing test_depend on 'rosunit'
sensor_msgs: notice: package description starts with boilerplate 'This package'
sensor_msgs: test/CMakeLists.txt(2): notice: use ${PROJECT_NAME} instead of 'sensor_msgs'
sensor_msgs: test/CMakeLists.txt(3): notice: use ${PROJECT_NAME} instead of 'sensor_msgs'
catkin_lint: checked 1 packages and found 5 problems
```

jack-oquin:
I'm inclined to ignore those notices.

The two errors look like preexisting problems, but it probably makes sense to fix them now, anyway.

k-okada:
@jack-oquin see https://github.com/ros/common_msgs/pull/103 for fix warning of catkin_lint

PR #105, opened 2017-05-22
Image/PointCloud Snapshot as a service

andreucm:
Some applications, such as industrial manipulation, do not require image/point_cloud streaming. Instead, they need image/point_cloud as a service, upon a request of some other client node, thus networking load is reduced, allowing other hardware to work better (specially that involving controllers).

This Pull Request defines image and point cloud snapshots as services. Examples of packages using these services can be found at: 
- [PtGrey BlackFly Monocular camera](https://github.com/beta-robots/ptgrey_bfly_camera)
- [Ensenso 3D camera ](https://github.com/beta-robots/ensenso_nx)

The Pull Request is against jade-devel branch because it is the latest one available, but the services have been tested on kinetic. However, it seems that they could work on jade and indigo.

### DIFF
--- a/sensor_msgs/srv/SnapshotCloud.srv
+++ b/sensor_msgs/srv/SnapshotCloud.srv
@@ -1,0 +1,14 @@
+#Request/Reply for Point Cloud as a service
+
+# Request:
+#   dense_cloud = true if the cloud has to be dense (ordered)
+#   exposure in milliseconds (0 indicates autoexposure) 
+bool dense_cloud
+uint32 exposure
+
+---
+
+# Reply:
+#   cloud is the returning point cloud
+sensor_msgs/PointCloud2 cloud 
+
--- a/sensor_msgs/srv/SnapshotCloud.srv
+++ b/sensor_msgs/srv/SnapshotCloud.srv
@@ -1,14 +1,17 @@
-#Request/Reply for Point Cloud as a service
+# Point Cloud as a service
 
-# Request:
-#   dense_cloud = true if the cloud has to be dense (ordered)
-#   exposure in milliseconds (0 indicates autoexposure) 
+# REQUEST:
+
+# dense_cloud = true, if the cloud has to be dense (ordered)
 bool dense_cloud
+
+# exposure in milliseconds (0 indicates autoexposure) 
 uint32 exposure
 
 ---
 
-# Reply:
-#   cloud is the returning point cloud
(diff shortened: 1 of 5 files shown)


sensor_msgs/srv/SnapshotCloud.srv
```diff
@@ -0,0 +1,17 @@
+# Point Cloud as a service
+
+# REQUEST:
+
+# dense_cloud = true, if the cloud has to be dense (ordered)
+bool dense_cloud
+
+# exposure in milliseconds (0 indicates autoexposure) 
+uint32 exposure
```
tfoote, line 9: Do you really expect to be setting the exposure time for a cloud? This seems to be breaking several abstractions that you must know about the sensor's parameters and lighting situation to request a snapshot?

andreucm: Do you mean to let the exposure config through another mechanism like, for instance, dynamic reconfigure, so keep the request even more simple ?

tfoote: Yes. This standalone argument means nothing unless you know a lot more about the system. It has implications for side effects too, such as will this modify the system if it's already running with the exposure setting? What does the exposure setting mean if this snapshot is implemented with a data source such as a laser?

sensor_msgs/srv/SnapshotImage.srv
```diff
@@ -0,0 +1,17 @@
+# Image as a service
+
+# REQUEST
+
+# exposure in milliseconds (0 indicates autoexposure) 
+uint32 exposure
```
tfoote, line 6: There's the same issue here. Unless you're going to provide all the settings providing just one doesn't make much sense.

andreucm: Yes, I agree. Providing one parameter is quite arbitrary and not general. I can modify both srv with an std_msgs/Empty in the request side. 
Another possibility could be to request with an std_msgs/UInt8 indicating the number of snapshots to take, and reply with a vector of either sensor_msgs/PointCloud2 or sensor_msgs/Image.
By the way, thanks for reviewing this PR!

tfoote (changes requested):
Marking this as changes requested so we know there's outstanding updates needed.

sloretz:
Thank you for the PR!

PR #106, opened 2017-06-14
Added test_depend on rosunit in sensor_msgs

Kukanani:
I noticed that this dependency was missing in sensor_msgs, which uses gtest. This is contrary to 
http://docs.ros.org/jade/api/catkin/html/howto/format1/gtest_configuration.html.

See also #85


tfoote:
Thanks!

PR #108, opened 2017-08-14
Support for dynamic PointCloud2 creation

spuetz:
This enhancement enables a dynamic creation of sensor_msgs::PointCloud2 messages with dynamic point fields at creation time of the cloud. This means a node or driver providing sensor_msgs::PointCloud2 clouds can now publish  configurable clouds. That is useful for devices which provide much extra information to each point in the cloud. With this enhancement, e.g. the user of a high resolution laser scanner, could configure which extra information should be published within the cloud. This is useful to only publish necessary information and to keep the network communication load smaller.

With the method `bool PointCloud2Modifier::addPointCloud2Field(std::string name, std::string datatype)` it is now possible to dynamically add fields if a field was enabled in the config.

### DIFF
--- a/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.h
+++ b/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.h
@@ -36,6 +36,9 @@
 #define SENSOR_MSGS_IMPL_POINT_CLOUD_ITERATOR_H
 
 #include <sensor_msgs/PointCloud2.h>
+#include <sensor_msgs/point_field_conversion.h>
+#include <sensor_msgs/point_cloud_conversion.h>
+
 #include <cstdarg>
 #include <string>
 #include <vector>
@@ -98,7 +101,7 @@ inline int addPointField(sensor_msgs::PointCloud2 &cloud_msg, const std::string 
 namespace sensor_msgs
 {
 
-inline PointCloud2Modifier::PointCloud2Modifier(PointCloud2& cloud_msg) : cloud_msg_(cloud_msg)
+inline PointCloud2Modifier::PointCloud2Modifier(PointCloud2& cloud_msg) : cloud_msg_(cloud_msg), current_offset_(0)
 {
 }
 
@@ -133,6 +136,7 @@ inline void PointCloud2Modifier::resize(size_t size)
 inline void PointCloud2Modifier::clear()
 {
   cloud_msg_.data.clear();
+  current_offset_ = 0;
 
   // Update height/width
   if (cloud_msg_.height == 1)
@@ -165,21 +169,21 @@ inline void PointCloud2Modifier::clear()
 inline void PointCloud2Modifier::setPointCloud2Fields(int n_fields, ...)
 {
   cloud_msg_.fields.clear();
+  current_offset_ = 0;
   cloud_msg_.fields.reserve(n_fields);
   va_list vl;
   va_start(vl, n_fields);
-  int offset = 0;
   for (int i = 0; i < n_fields; ++i) {
     // Create the corresponding PointField
     std::string name(va_arg(vl, char*));
     int count(va_arg(vl, int));
     int datatype(va_arg(vl, int));
-    offset = addPointField(cloud_msg_, name, count, datatype, offset);
+    current_offset_ = addPointField(cloud_msg_, name, count, datatype, current_offset_);
   }
   va_end(vl);
 
   // Resize the point cloud accordingly
-  cloud_msg_.point_step = offset;
+  cloud_msg_.point_step = current_offset_;
   cloud_msg_.row_step = cloud_msg_.width * cloud_msg_.point_step;
   cloud_msg_.data.resize(cloud_msg_.height * cloud_msg_.row_step);
 }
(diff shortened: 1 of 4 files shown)


spuetz:
Btw. the test units Jpr__common_msgs__ubuntu* seems to not allow `ROS_ERROR_**STREAM**("something" << "something")` so I'll change it to `ROS_ERROR()`.

sensor_msgs/include/sensor_msgs/point_field_conversion.h
```diff
@@ -80,6 +82,51 @@ namespace sensor_msgs{
   template<> struct typeAsPointFieldType<double>   { static const uint8_t value = sensor_msgs::PointField::FLOAT64; };
  
   /*!
+   * \Type names of the PointField data type.
+   */
+  int getPointFieldTypeFromString(const std::string field_name){
```
tfoote: Since this is not templated it needs to be inline with the implementation in the header. Also please switch to a const reference instead of pass by value of the string.

sensor_msgs/include/sensor_msgs/point_field_conversion.h
```diff
@@ -80,6 +82,51 @@ namespace sensor_msgs{
   template<> struct typeAsPointFieldType<double>   { static const uint8_t value = sensor_msgs::PointField::FLOAT64; };
  
   /*!
+   * \Type names of the PointField data type.
```
tfoote: Please add docblock info for the parameter like the other functions.

sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.h
```diff
@@ -197,36 +201,64 @@ inline void PointCloud2Modifier::setPointCloud2Fields(int n_fields, ...)
...
 
   // Resize the point cloud accordingly
-  cloud_msg_.point_step = offset;
+  cloud_msg_.point_step = current_offset_;
   cloud_msg_.row_step = cloud_msg_.width * cloud_msg_.point_step;
   cloud_msg_.data.resize(cloud_msg_.height * cloud_msg_.row_step);
 }
 
+
+inline void PointCloud2Modifier::addPointCloud2Fields(std::vector<PointFieldInfo> fields){
```
tfoote: Is there a reason this is void not at least bool?

sensor_msgs/include/sensor_msgs/point_cloud2_iterator.h
```diff
@@ -161,9 +161,23 @@ class PointCloud2Modifier
...
+  };
+
+  void addPointCloud2Fields(std::vector<PointFieldInfo> fields);
+  bool addPointCloud2Field(std::string name, std::string datatype);
+
 protected:
   /** A reference to the original sensor_msgs::PointCloud2 that we read */
   PointCloud2& cloud_msg_;
+    /** The current offset for all point fields */
+  int current_offset_;
```
tfoote: This seems to mirror the local storage of cloud_msg_.point_step ? Can you not query, mutate and reset without adding another data field?

sensor_msgs/include/sensor_msgs/point_cloud2_iterator.h
```diff
@@ -161,9 +161,23 @@ class PointCloud2Modifier
...
+
+  class PointFieldInfo{
+    public:
+      PointFieldInfo(std::string name, std::string datatype)
+         : name(name), datatype(datatype){}
+      std::string name;
+      std::string datatype;
+  };
+
+  void addPointCloud2Fields(std::vector<PointFieldInfo> fields);
```
tfoote: Please mirror  the setPointCloud2Fields API here instead of defining a new datatype.

tfoote (changes requested):
Overall adding this sort of functionality makes sense.

I had some specific comments inline. Overall all of these new functions need documentation at the level of doc blocks.

I'd also like to see the matched remove field methods. And looking at the methods it looks like the add methods will corrupt existing data in the point cloud. Some protections/warnings or non-destructive transformations would be best.

And to be confident this will work. Please add some unit tests that show it working.

spuetz:
@tfoote I changed the things you commented. The `PointCloud2Modifier::addPointCloud2Fields` as well as the whole `PointCloud2Modifier` will only work at creation time of a PointCloud2. Once the buffer is filled, the step size is fixed. That should be clear for the developers using that class.

Having the `PointCloud2Modifier::addPointCloud2Fields(const std::vector<PointFieldInfo>& fields)` method with a vector of `PointFieldInfo` will enable dynamic point field and point cloud creation at run time. For example you could switch on and off some fields in a laser scanner driver.

The next thing I want to do here is working on the `PointCloud2Iterator` to enable organized point structure access with something like a matrix access operation, e.g. `cloud(i, j)`, which is very useful for organized clouds to enable a fast access of neighboring points in the cloud. However, it would be very good to finish this pull request fast, so that I could start the next one. ;)

spuetz:
@tfoote

spuetz:
Hi @tfoote we would need the pull request now for the Sick MRS1000 Driver to dynamically enable a intensity channel in the point cloud.

tfoote:
Thanks for resolving the direct comments. However it still does not have any test coverage of the new API, they would also provide examples of how it's designed to be used which would be helpful in checking the design.

We're approaching Melodic release and getting this in before then would be great.

spuetz:
Okay I'll add one or more tests in the next days. Thank you. Would be could to have it in Melodic.

ctieben:
Are there any news about this PR?

spuetz:
We are planning on writing some tests soonish. @ctieben Do you need the feature?

ctieben:
I’m interested in this PR and looking forward to using these changes in some future experiments. It would be great to have this in the default branch / repos.

sloretz:
Thank you for the PR!

PR #109, opened 2017-09-03
Added equidistant distortion model const

alexzzhu:
Adds a new distortion model to the const definitions for fisheye cameras. Used in the following PR in vision_opencv/image_geometry to incorporate fisheye rectification into the standard pipeline:
https://github.com/ros-perception/vision_opencv/pull/184

### DIFF
--- a/sensor_msgs/include/sensor_msgs/distortion_models.h
+++ b/sensor_msgs/include/sensor_msgs/distortion_models.h
@@ -44,6 +44,7 @@ namespace sensor_msgs
   {
     const std::string PLUMB_BOB = "plumb_bob";
     const std::string RATIONAL_POLYNOMIAL = "rational_polynomial";
+    const std::string EQUIDISTANT = "equidistant";
   }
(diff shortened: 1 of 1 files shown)


tfoote:
This looks reasonable. Can you link to some more documentation on this distortion model?

alexzzhu:
The distortion model is described in the paper "A Generic Camera Model and Calibration Method for Conventional, Wide-Angle, and Fish-Eye Lenses" by Juho Kannala and Sami S. Brand ([paper](http://www.ee.oulu.fi/~jkannala/publications/tpami2006.pdf)), where it's referred to as the equidistance model (Kalibr refers to it as the equidistant model, hence the naming here). The model is used in OpenCV, [Kalibr](https://github.com/ethz-asl/kalibr/wiki), and the [MATLAB Camera Calibration Toolbox](http://www.vision.caltech.edu/bouguetj/calib_doc/index.html#links)

PR #112, opened 2017-09-14
Alternative approach for converting sensor_msgs/PointCloud2 to a list of named tuples 

2scholz:
This is an alternative approach to this [pull request](https://github.com/ros/common_msgs/pull/111). 
I implemented `read_points_list` using `read_points` and updated the documentation to make it clear that this function is slower. 

If this is merged we would like to see this cherry-picked to indigo.


tfoote (approved):
Thanks for iterating.

PR #117, opened 2018-01-31
Add polygon with 64bit precision

jgassaway:
There is no polygon with 64 bit precision, this adds a Polygon64 and Polygon64Stamped. 

Ideally, the standard naming conventions should be maintained, where Polygon.msg has 64-bit, and Polygon32.msg has 32-bit precision, but that would change the message definition and is more likely to break dependent code.

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -17,6 +17,8 @@ add_message_files(
   PointStamped.msg
   Polygon.msg
   PolygonStamped.msg
+  Polygon64.msg
+  Polygon64Stamped.msg
   Pose2D.msg
   Pose.msg
   PoseArray.msg
--- a/geometry_msgs/msg/Polygon64.msg
+++ b/geometry_msgs/msg/Polygon64.msg
@@ -1,0 +1,2 @@
+#A specification of a polygon where the first and last points are assumed to be connected (64 bits of precision)
+Point[] points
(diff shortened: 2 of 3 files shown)


tfoote:
Could you explain a little bit more about the use case where you see this as necessary? In general I'd prefer not to have duplicates of every message with different precisions for then every component dealing with that datatype needs to have dual parallel interfaces and possibly implementations. It might make more sense to have specific messages (with extra information) for a use case or cases that need the higher precision.

jgassaway:
For my part, I have a simple polygon drawing tool in mapviz. I am outputting a polygon in wgs84 (gps latitude and longitude) coordinates, where I actually need the precision to be low enough for centimeter accuracy or less.

tfoote:
For that use case a generic polygon 32 or 64bit is not appropriate. Latitude and longitude do not fit into Points (the units of points are meters). As a first pass I'd suggest simply using a vector of [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) datatypes, which natively represents latitude and longitude. There's clearly more information in there than you need. And a custom message could be created to be more specific.

What is the object that you're trying to represent? If you're defining a message it needs to capture the semantic meaning of the data so that it can be reused appropriately.

jgassaway:
Alright, sounds like I need a better approach for representing this data. I'll either create a different msg type, or handle my polygon creation another way. Thanks!

PR #125, opened 2018-05-03
Add Ray Message

Jmeyer1292:
My use case is publishing user mouse clicks in Rviz to external ROS nodes. Didn't find a standard message for a ray, so I here I am.

### DIFF
--- a/geometry_msgs/msg/Ray.msg
+++ b/geometry_msgs/msg/Ray.msg
@@ -1,0 +1,3 @@
+# A representation of a ray in 3-space, composed of an origin and a direction
+Point origin
+Vector3 direction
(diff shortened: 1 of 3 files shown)


tfoote:
This looks like a reasonable message definition. Can you point to places using this or a similar message for example use cases to show that in it's current form it will be useful. Are there equivalent datatypes in use in other geometry libraries? And do you have converter and transform operations that could be added to tf2_geometry_msgs to support it?

geometry_msgs/msg/Ray.msg
```diff
@@ -0,0 +1,3 @@
+# A representation of a ray in 3-space, composed of an origin and a direction
+Point origin
+Vector3 direction
```
tfoote, line 3:
```suggestion
Quaternion direction
```

A direction cannot have magnitude, but a Vector3 does.

sloretz:
Thank you for the PR!

PR #128, opened 2018-07-04
included missing import for the read_points_list method

jgueldenstein:
The read_points_list method is currently broken since the import of namedtuple is missing.

Please include this in all current branches

@v4hn

### DIFF
--- a/sensor_msgs/src/sensor_msgs/point_cloud2.py
+++ b/sensor_msgs/src/sensor_msgs/point_cloud2.py
@@ -40,13 +40,13 @@ Serialization of sensor_msgs.PointCloud2 messages.
 Author: Tim Field
 """
 
+from collections import namedtuple
 import ctypes
 import math
 import struct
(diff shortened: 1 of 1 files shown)


sensor_msgs/src/sensor_msgs/point_cloud2.py
```diff
@@ -46,6 +46,7 @@
 
 import roslib.message
 from sensor_msgs.msg import PointCloud2, PointField
+from collections import namedtuple
```
tfoote: Could you move this above into the system imports in alphabetical order?

tfoote (commented):
Thanks for fixing this. One formatting request and then it looks good to go.

And yes this should be mergeable back into indigo too.

v4hn:
@tfoote can we get this released back in kinetic soon?
I have a practical course coming up and would like students to rely on this fix...

PR #129, opened 2018-07-06
feat(pose2d_stamped): Added Pose2D Stamped message

reinzor:


tfoote:
Could you please provide some context as to why you'd like to add this message?

reinzor:
For some of our local planners, we would like to expose references or errors. The pose stamped message would suffice however it is all 2d and is is not easy to plot or view the yaw error for example.

tfoote:
In general our recommendation is to use a full 3D representation of everything and for 2D specific applications make the appropriate projections into the plane for their calculations but optimally will preserve the 3D information during processing.

If we have parallel copies of 2D datatypes every UI and other pipeline will end up needing to have dual interfaces to plot everything. And you will end up with not being able to use 3D tools for 2D use cases even if they're completely valid, as you'd have to reimplement it with different inputs and outputs. It's not particularly hard to plot the 2D pose or compute the yaw error for the Pose message and there are already tools and libraries that can do this for you.

As an example of where it's powerful, if you have a camera detected goal object. You may want the base planner to navigate to next to the goal, but then you'll want to reach for the goal with your robot's arm. Being able to have all tools process the 3D goal and derivative poses is much simpler if everything operates on the full 3D datatypes.

reinzor:
Thanks for your elaborate comment. I do agree after reading it; It might be better way then to improve the plot GUI functionalities. Can I ask one last question. Why is the Pose2D (without stamped) present in the geometry_msgs, or is this just legacy?

tfoote:
Yeah, it's still there for legacy reasons. Early on we created a bunch of messages on speculation for "completeness". But since then we've learned not to do that. I should add a note in the documentation related to this, it's a not uncommon question.

Here's the initial commit talking about it being "unused"
https://github.com/ros/common_msgs/commit/4b4d7905f85851a153448af774308dff8f7d2c40#diff-6c097b109f04f06d9ad03d3eced1b5a9

tfoote:
PR in #130 to add a comment in the Pose2D message.

Also contributions for plotting the Pose in GUIs would definitely be appreciated.

PR #130, opened 2018-07-11
Add deprecation comment about Pose2D

tfoote:
It started as unused and still isn't recommended.

Followup to #129

### DIFF
--- a/geometry_msgs/msg/Pose2D.msg
+++ b/geometry_msgs/msg/Pose2D.msg
@@ -1,3 +1,11 @@
+# Deprecated
+# Please use the full 3D pose.
+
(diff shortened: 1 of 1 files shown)


130s:
I think I agree with this direction. I see opposing comment at [answers.ros.org#q339764](https://answers.ros.org/question/339764/why-geometry_msgspose2d-in-ros2-is-not-deprecated/) from @SteveMacenski, so I'm just interested in how the conversation would develop.

DLu:
I'm very late, but I disagree with this move.

SteveMacenski:
Agreed

ros-discourse:

https://discourse.ros.org/t/discussion-standardizing-ros-interfaces/33693/1

PR #131, opened 2018-08-15
Include sstream on header that needs it

chapulina:
Similar to https://github.com/ros2/common_interfaces/pull/54, targeted at ROS2


tfoote (approved):
yup, that was missing.

PR #140, opened 2018-12-07
Update BatteryState.msg with temperatures

### LINKED ISSUE #139: Add cell_temperature to sensor_msgs/BatteryState
Often the temperatures are also measured but cannot be put in the BatteryState message.

reinzor:
Resolves https://github.com/ros/common_msgs/issues/139

### DIFF
--- a/sensor_msgs/msg/BatteryState.msg
+++ b/sensor_msgs/msg/BatteryState.msg
@@ -33,6 +33,7 @@ uint8 POWER_SUPPLY_TECHNOLOGY_LIMN = 6
 
 Header  header
 float32 voltage          # Voltage in Volts (Mandatory)
+float32 temperature      # Temperature in Degrees Celsius (If unmeasured NaN)
 float32 current          # Negative when discharging (A)  (If unmeasured NaN)
 float32 charge           # Current charge in Ah  (If unmeasured NaN)
 float32 capacity         # Capacity in Ah (last full capacity)  (If unmeasured NaN)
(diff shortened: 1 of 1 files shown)


reinzor:
ping

tfoote:
This looks like a reasonable addition. Changing a core message like this is not something that we can do during a release cycle so we will need to target Noetic. And it would be good to make sure to clearly document any migration requirements for existing users, both for requirements for their code base as well as bag migration rules for logged data.

clalancette (approved):
This looks good to me, provided that we create a new "noetic-devel" branch and target that.  We also want this in ROS 2; I'll make a separate PR for that.

reinzor:
Can this be closed or do you still require this?

clalancette:
I think this is a good idea to do; thanks for pinging.  Now that we are heading towards noetic, I'd create a new branch (noetic-devel), retarget this there, and then merge it.  @tfoote any thoughts?  If not, I'll just go ahead and do it.

clalancette:
All right, I've created a new noetic-devel branch, then retargeted this PR at that branch.  I'm going to go ahead and merge it now.  We'll also need an update to https://github.com/ros/rosdistro/blob/master/noetic/distribution.yaml to update the branches there, which I'll do shortly.

reinzor:
Thanks!

tfoote:
Branch change upstream for cross reference. https://github.com/ros/rosdistro/pull/24094

dirk-thomas:
This has not been released into Noetic. Doing so after the release breaks the interface.

gavanderhoorn:
Just to keep things connected: this PR has come up on ROS Answers: [How to receive standard message that has changed format in noetic?](https://answers.ros.org/question/363757).

PR #141, opened 2018-12-25
Add PolygonArray.msg

ShepelIlya:
Added PolygonArray.msg. My usecase is publishing several different polygons with one timestamp. Didn't find any standard way to do it.


tfoote:
What is the semantic meaning of a PolygonArray with a header? In general this looks like a good candidate for you to define a custom message that has semantic meaning instead of just creating a message with a header and an array of other elements.

ShepelIlya:
Ok. I'll create my own message. I considered that each geometry message must have an analog with an array, because must of all others messages have ...Array.msg.

PR #147, opened 2019-10-09
Added AnchorScan.msg

elcingencturk:
This is a message to hold the fixed anchors' (UWB sensors) information at known coordinates and Time Difference of Arrival (TDOA) values. The indoor positioning systems developed using UWB signals include multiple sensors that transmit UWB signals in an environment. Calculations such as position, anchor selection etc. are made by using TDOA measurements from these sensors and coordinate information of the sensors.  With the defined AnchorScan message, it allows the publish of this information from the sensors via any developed firmware.

### DIFF
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -11,6 +11,7 @@ include_directories(include)
 add_message_files(
   DIRECTORY msg
   FILES
+  AnchorScan.msg
   BatteryState.msg
   CameraInfo.msg
   ChannelFloat32.msg


tfoote (changes requested):
This is failing CI because the patch does not have the message definition. Likely it wasn't added before the commit was made. 

However even prior to reviewing the message itself this seems moderately early in the development cycle to be merged into common messages. In general common_msgs is where we will put messages that are mature and well tested and maybe are a small iteration on existing messages or are a merge of a couple of different implementations that have traction but had differences that needed to be resolved.

To that end I'd suggest that this be not immediately released here in sensor_msgs but be released into something more like `uwb_msgs` or maybe something a little bit more generic for beacon based localization messages. That then can be developed against and potentially iterate more rapidly than we can do in common_msgs.

sloretz:
Thank you for the PR!

PR #148, opened 2019-10-09
Create AnchorScan.msg

elcingencturk:
This is a message to hold the fixed anchors' (UWB sensors) information at known coordinates and Time Difference of Arrival (TDOA) values. The indoor positioning systems developed using UWB signals include multiple sensors that transmit UWB signals in an environment. Calculations such as position, anchor selection etc. are made by using TDOA measurements from these sensors and coordinate information of the sensors. With the defined AnchorScan message, it allows the publish of this information from the sensors via any developed firmware.


tfoote:
This appears to be the second half of #147 please merge these changes into that branch and continue the conversation there. 

Looking at this message reinforces my response in that review however that this message does not look mature enough to be promoted to common_msgs and could use some iterations.

PR #152, opened 2019-11-25
Add LoadMap service definition

jacobperron:
Continuation of https://github.com/ros/common_msgs/pull/95. Now targeting the default branch.

---

Original description:

The service is for loading a map as an [nav_msgs/OccupancyGrid](http://docs.ros.org/indigo/api/nav_msgs/html/msg/OccupancyGrid.html) given a resource ID. It also features a **result** member to encode errors. I tried to make the result codes generic enough for support of arbitrary backends and underlying data structures.   

This could be implemented in a *map_loader* node or directly in [map_server](https://github.com/ros-planning/navigation) seeing how it already "loads" a map on startup.

---

I've updated the definition to take a URL as suggested by @jack-oquin (774d96f) and fixed an error (a12d806).


jacobperron:
> If you merge this for Eloquent, I'll change my Nav2 PR to be based on this service.

@mkhansen-intel This PR is targeting ROS 1. But if it is approved and merged, I'll port it to ROS 2.
Since this may take some time, I would go ahead with the nav2 version of the service definition and we can make the change to the common type in the future. Hopefully it's as easy as swapping the package name :crossed_fingers:

jacobperron:
> So, do we now require that every image is a full path?

Good question. I figured that the path to image file is assumed relative to the location of the YAML file. I'd say no, we shouldn't use URLs for the image files. I think we should be compatible with whatever map saver is doing.

nav_msgs/srv/LoadMap.srv
```diff
@@ -0,0 +1,15 @@
+# URL of map resource
+# Can be an absolute path to a file: file:///path/to/maps/floor1.yaml
+# Or, relative to a ROS package: package://my_ros_package/maps/floor2.yaml
+string map_url
```
jacobperron, line 4: I'd like to posit that **how** the map is loaded or **what** type of file the URL points to is up to the node implementing the service. E.g. it might point to a yaml file containing the metadata, or maybe an image file directly.

See https://github.com/ros-planning/navigation2/pull/1303#discussion_r352957521 for a relevant discussion.

cc @SteveMacenski

SteveMacenski: @jacobperron this is also for ROS1, I'm not as concerned about it.

However you think we can change this to `map_uri`? I typically think of URLs as links to a server rather than a URI which can be a local path.

nav_msgs/srv/LoadMap.srv
```diff
@@ -0,0 +1,15 @@
...
+---
+# Result code defintions
+uint8 RESULT_SUCCESS=0
```
SteveMacenski, line 7: I'd prefer if these were just `SUCCESS` etc without the `RESULT_` prefix since in use it would be `nav_msgs::msg::LoadMap::Result::RESULT_MAP_DOES_NOT_EXIST` and we see `result` twice in a row.

mkhansenbot: Actually, its not:
`nav_msgs::msg::LoadMap::Result::RESULT_MAP_DOES_NOT_EXIST`
It's
`nav_msgs::msg::LoadMap::Response::RESULT_MAP_DOES_NOT_EXIST`

Subtle difference.

nav_msgs/srv/LoadMap.srv
jacobperron: A URL can also be a local path. It describes a way to **locate** a resource.
I'm okay with changing this to a more general URI, but would like to hear what others think. It might make the semantics more ambiguous.

SteveMacenski: I know a URL 'is' that, but in vernacular we usually talk about servers. A URL is a subclass of URI. 100% agree its semantics. I just think it makes it more clear actually because of the vernacular uses and URI is what is used in SDF for getting models/packages/files in the same way this is attempting to so its consistent with other ROS-y tools.

I don't care that much one way or another since I know what it means, but I could see someone just looking at a message definition in `ros2 msg show ...` and thinking it _had_ to be a web link. If no one else agrees, I'm not going to push it.

nav_msgs/srv/LoadMap.srv
SteveMacenski: ah, yes sorry. 

Its not something I care deeply about, just a comment that it seems related enough with the namespace the additional wording just uses linespace. If no one else agrees, I'm certainly not going to push it.

SteveMacenski:
I believe we're using the Nav2 msgs for now on this: https://github.com/ros-planning/navigation2/blob/master/nav2_msgs/srv/LoadMap.srv

I'm not too concerned personally with this being merged (at this point, I think the PR should be closed). In the medium term future, I want nav2_msgs put into nav_msgs, but I want to wait until the APIs settle down. We're still working on adding result and feedback types to all actions.

jacobperron:
I think @tfoote is the maintainer of this package and should be the one to review / merge.

SteveMacenski:
That's also fine.

DLu:
Alternative: We could put it in [`map_msgs`](https://github.com/ros-planning/navigation_msgs/tree/ros1/map_msgs/srv)

SteveMacenski:
Also fine with that.

SteveMacenski:
@jacobperron what's the blocker from this being merged? It looks like Matt, David, and I all approve it. I don't know if David has admin here to merge it but I don't. What's the reason this isn't in / does this just need someone to hit the button?

jacobperron:
@SteveMacenski I'm not the maintainer of this package. I believe the responsibility lies on @tfoote.

tfoote (approved):
We can add this here. Especially since it's been proven out in the nav2_msgs already. 

It's purely an addition so it will be backwards compatible. What distributions are most important to target? By default I'm going to retarget this to noetic-devel to merge.

SteveMacenski:
Throw it into Foxy as well, I'll file a ticket then to migrate to this message uses' over `nav2_msgs`

DLu:
This PR is for the `jade-devel` branch, but confusingly, that branch is used for kinetic and melodic, so merging this should update both those distros. I've cherry-picked the commits for noetic in #164

SteveMacenski:
Is there a potential to have an analog for ROS2?

jacobperron:
If you propose a PR to the default branch of https://github.com/ros2/common_interfaces I can review it.

SteveMacenski:
https://github.com/ros2/common_interfaces/pull/129 done

PR #153, opened 2019-12-07
Add Microwave radiation type.

rikba:
This PR adds `MICROWAVE` as a radiation type to range sensors. For example to capture radar altimeters such as [uLanding](https://aerotenna.readme.io/docs/what-is-%CE%BClanding).

### DIFF
--- a/sensor_msgs/msg/Range.msg
+++ b/sensor_msgs/msg/Range.msg
@@ -15,6 +15,7 @@ Header header           # timestamp in the header is the time the ranger
 # If you want a value added to this list, send an email to the ros-users list
 uint8 ULTRASOUND=0
 uint8 INFRARED=1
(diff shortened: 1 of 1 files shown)


rikba:
Note that the uLanding altimeter is deprecated and the new version is called [US-D1](https://ainstein.ai/drone-makers-drone-service-providers/us-d1/). But the term microwave still applies.

sloretz:
Thank you for the PR!

tfoote:
As a note though this could be reopened against https://github.com/ros2/common_interfaces/blob/edb0a3a9ed3c8309aada12e9ced9a65e532cf1a1/sensor_msgs/msg/Range.msg#L17 for inclusion in ROS Rolling and future releases.

PR #154, opened 2019-12-12
Add feedback to GetMap.action

JaccovdS:
The `GetMap.action` now only contains a result. This PR extends the action with feedback which is useful for getting progress updates while doing mapping.

### DIFF
--- a/nav_msgs/action/GetMap.action
+++ b/nav_msgs/action/GetMap.action
@@ -2,4 +2,4 @@
 ---
 nav_msgs/OccupancyGrid map
 ---
-# no feedback
+nav_msgs/OccupancyGrid progress_map


tfoote:
Can you please explain what a `progress_map` is and how does it indicate progress on a getter type of call?

JaccovdS:
The progress map is basically a partial map that has been build up during the mapping process until now. I.e. if you are mapping a big space you also might want to track the progress visually. Some packages now provide a topic to which you can subscribe to see the map created until now. However since mapping can also be considered as a process which takes a long time it makes sense to provide an action interface which not only provides the resulting map but also the 'in between' versions.

tfoote:
As I understand this action it's a non-blocking analog to the service GetMap.srv and should return the current map as soon as practical. I'm not familiar with any general concept of when a map is complete. And this action does not therefore expect to wait for "completeness". There is some point when a human often says "good enough" and stops building/updating the map. But that's more a practical thing that we don't have standard implementations of life long mapping where maps are continuously updating with changes to the map. Which again comes back to the lack of concept of completeness. Similarly if this is a feedback message for the sense of progress it should convey some information which provides feedback on the process of the action which is GetMap. 

It sounds more like you're implementing an action which I would classify as exploring the environment and presumably implementing an exploration algorithm which will explore until you have coverage of all the frontiers are explore or some other metric. To get to that end you should create a specific action to cover that semantic use case. And then you can add in meaningful progress indicators such as a hypothetical frontier coverage ratio and if you want it could also stream back the in progress map. But you seem to be adding a lot more semantic meaning and that should be caputred into your action definition and not try to shoe horn it into an existing generic action like GetMap.

JaccovdS:
I agree with you that there is no `completeness` in a mapping process in general. And that is the reason that I think it would be beneficial to provide the map i.e. the `OccupancyGrid` that has been built up until this point in time as feedback. The user can use for example `rviz` to judge if exploration should be terminated by monitoring this feedback. 

I agree with you that there might be situations where you want to add additional or different information to cover different use-cases and these should be captured in a different action message. However I don't see why that is a reason to not add feedback to this action. The action does not become less generic by adding this feedback.

tfoote (changes requested):
Being generic is not the goal of these messages. If there's not a specific semantic meaning then it will cause issues with reuse. 

If you want to know the status of the map as you're progressing then, a good solution is to either publish the map periodically or to call GetMap periodically. In general viewing the map in rviz is something that's already made available by default. There's no need to stream a second copy of it only during this GetMap request.

The way you're proposing to structure this is not getting feedback on the action in question "GetMap" but on a larger system "BuildMap" that has valid semantics but they should not be added to this simple scoped action of GetMap. A new action with the specific semantics for your application should be designed for that use case.

JaccovdS:
Change doesn't seem to be inline with the philosophy. Closing PR.

PR #155, opened 2019-12-13
Fix TabError: inconsistent use of tabs and spaces in indentation

Rayman:
Python 3 is much more strict for spacing. I've replaced all 8
consecutive spaces with a tab. I did not touch the spacing inside the
messages as they are part of the message.


tfoote (changes requested):
Our standard is to use spaces and not tabs in the ROS project. http://wiki.ros.org/PyStyleGuide please convert it the other way to remove the tabs.

Rayman:
I've replaces all tabs with spaces according to PEP 8. Note that the diff is now bigger. I've just tested that this PR still fixes the `rosbag check` inconsistent spaces error.

tfoote (approved):
Thanks for iterating it looks good to me.

Issue #126, opened 2018-06-14
Create_cloud does not respect point cloud organization


haudren:
I was playing around with a PCL application that needs an organized point cloud as input. However, using `do_transform_point_cloud` (from `tf2_sensor_msgs`) returns an unorganized point cloud.

After a quick investigation. it seems like `create_cloud` always returns a 1 by `n` cloud, but I'm not sure if it would be enough to simply pass in width / height arguments or if the point packing must be adapted.

Hi-Zed:
I created a pull request ( #157) to fix this.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #157, opened 2020-01-20
Create ordered point cloud (issue #126)

Hi-Zed:
Small changes to create_cloud in point_cloud2.py to support ordered point clouds with specific height and width.
Mentioned in issue #126

### DIFF
--- a/sensor_msgs/src/sensor_msgs/point_cloud2.py
+++ b/sensor_msgs/src/sensor_msgs/point_cloud2.py
@@ -157,19 +157,18 @@ def create_cloud(header, fields, points, height=1, width=None):
     @return: The point cloud.
     @rtype:  L{sensor_msgs.msg.PointCloud2}
     """
-    if height == 1:
+    if height == 1 or width is None:
+        height = 1
         width = len(points)
 
     cloud_struct = struct.Struct(_get_struct_fmt(False, fields))
 
     buff = ctypes.create_string_buffer(cloud_struct.size * len(points))
 
-    point_step = cloud_struct.size
-    pack_into = cloud_struct.pack_into
     offset = 0
     for p in points:
-        pack_into(buff, offset, *p)
-        offset += point_step
+        cloud_struct.pack_into(buff, offset, *p)
+        offset += cloud_struct.size
 
     return PointCloud2(header=header,
                        height=height,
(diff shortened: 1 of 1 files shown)


haudren:
Thanks for the patch, it looks straightforward and a good addition. I hope this can get merged!

sensor_msgs/src/sensor_msgs/point_cloud2.py
```diff
@@ -150,28 +150,35 @@ def create_cloud(header, fields, points):
...
                    elements of each iterable being the values of the fields for 
                    that point (in the same order as the fields parameter)
+    @param height: the height of the point cloud [default: 1]
+    @type height: unsigned int
+    @param width: the width of the point cloud [default: None]
+    @param width: unsigned int
     @return: The point cloud.
     @rtype:  L{sensor_msgs.msg.PointCloud2}
     """
+    if height == 1:
```
tfoote: This looks like it will have problems if height is non unary but width is defaulted to None. Some other checks are going to be necessary.

sensor_msgs/src/sensor_msgs/point_cloud2.py
```diff
@@ -150,28 +150,35 @@ def create_cloud(header, fields, points):
...
+        width = len(points)
 
     cloud_struct = struct.Struct(_get_struct_fmt(False, fields))
 
     buff = ctypes.create_string_buffer(cloud_struct.size * len(points))
 
-    point_step, pack_into = cloud_struct.size, cloud_struct.pack_into
+    point_step = cloud_struct.size
```
tfoote: These intermediate values appear to only be used once so are probably not worth creating at all.

tfoote (changes requested):
This looks reasonable. Though it needs a little bit more error handling. None times an int will except.

sensor_msgs/src/sensor_msgs/point_cloud2.py
Hi-Zed: I'll add a check on both height and width.
The original implementation didn't perform any check on header, fields and points, for example, if they are empty. Do you want me to add something like that?

tfoote: I was mostly worried about the default values messing things up. Though a little bit of sanitizing would be valuable. But we need to be careful about what error response behavior is caused by invalid data.

tfoote (commented):
Thanks for the update that looks better. Could you also write some basic unit tests. It would be good to add some that assert the ordering that this implements stays valid. As well as can validate the error handling. And it appears that we don't have any python unit tests for this at all.

sloretz:
Thank you for the PR!

PR #158, opened 2020-01-25
Bump CMake version to avoid CMP0048 warning

sloretz:
This bumps the minimum CMake version to 3.0.2, which is the [minimum supported by ROS Kinetic](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021) and new enough to default to the NEW behavior of [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html). This avoids a CMake warning when building this package in Debian Buster and Ubuntu Focal.

Same as ros/catkin#1052


tfoote (approved):
Thanks

dirk-thomas:
This has not been released into Noetic so building from source after fetching the released code still results in a warning.

PR #166, opened 2020-09-02
YUV420 support

luca-della-vedova:
This PR adds support for YUV420 (or I420) images.

I420 is the most common YUV format and is adopted by most Image Sensor Processor (ISP) chips, specifically this was implemented to work with [gstreamer based v4l plugins](https://answers.ros.org/question/343864/how-can-i-publish-i420-image-data-as-sensor_msgsimage-image-transport-plugin/?answer=360789#post-id-360789)

I420 consists in 12 bits per pixel, which is implemented in OpenCV as a cv::Mat with one 8 bit channel with 1.5x the number of rows of the actual image.


sloretz:
Thank you for the PR!

PR #168, opened 2020-10-07
Update package maintainers.

hidmic:
Precisely what the title says.

PR #170, opened 2021-01-11
[melodic] Update package maintainers.

hidmic:
Same as #168, but for Melodic and other pre-Noetic distros going back to Jade.


clalancette (approved):

hidmic:
Thanks for the review!

PR #172, opened 2021-04-08
Add ImageMarkerArray message

amacneil:
Proposal to add `ImageMarkerArray`, which is to `ImageMarker` as `MarkerArray` is to `Marker`.

Can also port this to ros2 fairly easily if needed.

### DIFF
--- a/visualization_msgs/CMakeLists.txt
+++ b/visualization_msgs/CMakeLists.txt
@@ -7,6 +7,7 @@ add_message_files(
   DIRECTORY msg
   FILES
   ImageMarker.msg
+  ImageMarkerArray.msg
   InteractiveMarker.msg
   InteractiveMarkerControl.msg
   InteractiveMarkerFeedback.msg
--- a/visualization_msgs/mainpage.dox
+++ b/visualization_msgs/mainpage.dox
@@ -8,6 +8,7 @@ The messages it currently contains are:
  - \ref visualization_msgs::Marker
  - \ref visualization_msgs::MarkerArray
  - \ref visualization_msgs::ImageMarker
+ - \ref visualization_msgs::ImageMarkerArray
  - \ref visualization_msgs::Polyline
 
 */
--- a/visualization_msgs/msg/ImageMarkerArray.msg
+++ b/visualization_msgs/msg/ImageMarkerArray.msg
@@ -1,0 +1,1 @@
+ImageMarker[] markers


tfoote:
What use case does this serve and what semantic meaning does it provide? This does not appear to add any semantic meaning which means that it's hard to reuse effectively as a standalone. And being just an array it can be embedded as a submessage just as easily as `ImageMarker []` versus `ImageMarkerArray`. 

To make this message valuable it should have non-trivial documenatation which defines how to interpret the data. In particular what is the expected semantic meaning of the grouping.

elephantum:
+1 for adding ImageMarkerArray

My use case is to send results of ML-detection model as a bunch of markers to topic instead of rendering results to image.
Foxglove Studio offers ability to render marker topic on image topic, hence much more effective client side rendering.

Array is needed in order to render more than one marker at once.

Pre-defined (vs custom msg) is needed because Foxglove does not know anything about my custom msg type, so it can't use it for rendering.

amacneil:
Exactly. `ImageMarker` provides a way to publish image annotations on a separate topic (so that they can be rendered separately by viewing software). All the semantic meaning exists within `ImageMarker`.

`ImageMarkerArray` is simply a way for users to publish multiple annotations in a single frame (for example detected objects). It is useful for the same reason `MarkerArray` is useful for rviz.

It would be helpful to include this as a standard message definition in ROS so that libraries and tools can be interoperable.

jhurliman:
Until this is merged, I've published this as `foxglove_msgs/ImageMarkerArray` - https://github.com/ros/rosdistro/pull/29682

amacneil:
Closing PR because:
1. No further major releases of ROS 1 are planned after noetic. Any new messages should go in [ros2/common_interfaces](https://github.com/ros2/common_interfaces).
2. No "official" ROS visualization tools (rviz, rqt) even use the `ImageMarker` message - as far as I know Foxglove Studio and Webviz are the only tools that can render it (see: [ros answers](https://answers.ros.org/question/11982/are-there-any-viewers-for-imagemarker/)). So I don't see much point in adding this to ROS 2 unless other tools are going to use it as well.

Since image markers are a Foxglove-specific feature for now, we have chosen to publish [foxglove_msgs/ImageAnnotations](https://foxglove.dev/docs/studio/messages/image-annotations) under a separate package for ROS 1 and 2.

PR #177, opened 2022-03-22
Created TransformWithCovariance and TransformWithCovarianceStamped.

RobertBlakeAnderson:
Created covarianced versions of the transform messages. This was motivated by an application in which our system tries to estimate the poses of two objects relative to each other, in which we are interested in the uncertainty of that estimate.

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -26,7 +26,9 @@ add_message_files(
   Quaternion.msg
   QuaternionStamped.msg
   Transform.msg
+  TransformWithCovariance.msg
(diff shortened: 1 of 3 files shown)


tfoote:
This seems reasonable. However as this repo is targeted at ROS 1 distros which are all stabilized now and this would be better to target for future distributions in [ros2/common_interfaces](https://github.com/ros2/common_interfaces) Please repropose this addition there.

When you do repropose it there please consider providing examples of your usage. In particular I'd like to understand how you're using the frame_ids of the target objects your estimating versus a transform providing information between coordinate frames as we usually do.

PR #189, opened 2023-03-09
add: Vector3WithCovariance, Vector3WithCovarianceStamped

Masa0u0:
@tfoote 
I am going to use these messages to represent the linear velocity obtained from GPS.
There must be many other usages.

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -33,6 +33,8 @@ add_message_files(
   TwistWithCovarianceStamped.msg
   Vector3.msg
   Vector3Stamped.msg
+  Vector3WithCovariance.msg
+  Vector3WithCovarianceStamped.msg
   Wrench.msg
   WrenchStamped.msg)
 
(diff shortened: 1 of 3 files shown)


Masa0u0:
@tfoote

tfoote (changes requested):
Thanks for proposing this. However you specifically are asking for a generic message without semantic meaning. Generic messages like that should not be used standalone but should used only as components to construct a semantically meaningful datatype. 

This message could easily represent both linear velocity or position. These semantic value should be defined in the message. Telling you what the units are and preventing semantically invalid implicit type conversions. 

This also goes against our default policy to use the full degrees of freedom such that we don't need to write conditional logic to operate on partial information (aka upconvert LinearVelocity to Twist with Angular velocity zeroed out to do an operation, and then downsample it again.) For example what happens if your operation causes rotation while doing the math? That output would have to be discarded. So I would strongly recommend just using the Twist* messages and keep the angular velocities zeroed.

Masa0u0:
Thank you very much. I totally understand.
Then I will use TwistWithCovarianceStamped to express GPS velocity.

PR #190, opened 2023-03-21
Adds ARROW_STRIP Marker type

riv-tnoble:
Adds a new `ARROW_STRIP` `Marker` type for rendering batches of arrows.

Rviz Issue: https://github.com/ros-visualization/rviz/issues/1785
Rviz PR: https://github.com/ros-visualization/rviz/pull/1786


riv-tnoble:
Hey @tfoote, could I get you to take a look at this? The corresponding Rviz PR is ready for review

visualization_msgs/msg/Marker.msg
```diff
@@ -12,6 +12,7 @@ uint8 POINTS=8
 uint8 TEXT_VIEW_FACING=9
 uint8 MESH_RESOURCE=10
 uint8 TRIANGLE_LIST=11
+uint8 ARROW_STRIP=12
```
tfoote, line 15: Would it make more sense to call this a DIRECTED_LINE_STRIP as that's an already determined thing and we're just adding direction. 

There's a few places below that reference LINE_STRIP that should be augmented.

tfoote (commented):
Overall this seems like a good addition. My first reaction is to suggest that it be targeted for https://github.com/ros2/common_interfaces not here. This is a bit of a feature enhancement going into a stable release. But this is small enough that I would consider making it an exception. 

However I would like to see it reviewed and merged to the new development version: https://github.com/ros2/common_interfaces and then once that's reviewed backported here. That way anyone who starts using it here won't get caught flat footed when they want to move forward.

visualization_msgs/msg/Marker.msg
riv-tnoble: The geometry shares some functionality with `ARROW`, most notably the appearance and semantics of the `scale` field. I think `ARROW_STRIP` makes this connection more explicit, and conveys the same information as `DIRECTED_LINE_STRIP` in fewer characters.

riv-tnoble:
> Overall this seems like a good addition. My first reaction is to suggest that it be targeted for https://github.com/ros2/common_interfaces not here. This is a bit of a feature enhancement going into a stable release. But this is small enough that I would consider making it an exception.
> 
> However I would like to see it reviewed and merged to the new development version: https://github.com/ros2/common_interfaces and then once that's reviewed backported here. That way anyone who starts using it here won't get caught flat footed when they want to move forward.

That makes sense. The reason for targeting ROS1 was simply that it's the platform we're currently using, though we're planning to migrate to ROS2 next month. I'd be happy to raise the PR in https://github.com/ros2/common_interfaces, although due to software restrictions on my work machine, I likely won't be able to work on a ROS2 version of the corresponding Rviz PR until we start the migration in April.

I'll raise that PR now :+1:

riv-tnoble:
Closed in favour of https://github.com/ros2/common_interfaces/pull/218

tfoote:
Thanks for jumping it over there.

